PR #429, opened 2015-01-07
Handle whitespace URLs

willkg:
If a URL consists solely of whitespace, we should strip the whitespace
and store an empty string in the db.

r?

### DIFF
--- a/fjord/feedback/tests/test_views.py
+++ b/fjord/feedback/tests/test_views.py
@@ -514,6 +514,17 @@ class TestFeedback(TestCase):
         feedback = models.Response.objects.latest(field_name='id')
         eq_(feedback.url, u'http://mozilla.org/path/')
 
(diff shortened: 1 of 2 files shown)


willkg:
This removes leading/trailing whitespace from the data and then passes it through `clean_url`. The result being that if the url is just whitespace, it'll get saved as an empty string (which is good). Plus we ditch leading/trailing whitespace if it shows up, too.

To review:

1. look through the code changes and test
2. run the tests

lgp171188:
+1

lgp171188:
:+1:

rlr:
r+ too

willkg:
^^^ Squashed those two commits.

willkg:
Landed in: 43ec5da Handle whitespace URLs

Thank you!

PR #430, opened 2015-01-07
Fix commit message lint to gracefully handle empty commit message

lgp171188:
When the commit message is empty (for example when the developer wants to abort the commit), the commit message lint script throws an exception.

Now the commit message lint gracefully handles this scenario without throwing any errors or exceptions.


willkg:
Looks good to me! r+

willkg:
0aa0245 Fix commit message lint to gracefully handle empty commit message

PR #431, opened 2015-01-08
Remove analytics_flagged view.

rlr:
This seems to be leftover from the spicedham experiment.
It 500s.

r?


willkg:
I need to think about this. I think there was a reason I left this in, but I'll have to go back in time to prior me to find out if there was or not.

willkg:
@rlr showed how this is busted. +1 for removing it.

We should remove the `fjord/analytics/templates/analytics/analyzers/flags.html` template, too. Then this is good to go.

rlr:
Removed the template too :point_up:

willkg:
Looks good! r+

willkg:
2e006e9 Remove fjord/analytics/templates/analytics/analyzer/flags.html
74c3ed8 Remove analytics_flagged view.

PR #433, opened 2015-01-08
[bug 1093222] Remove django-cache-machine 

rlr:
This is it. I grep'd for `uncached` and `no_cache` and `caching.base` and am pretty sure I got everything.

Unit tests pass.
Smoke tests pass.
What could go wrong?

r?

### DIFF
--- a/fjord/base/models.py
+++ b/fjord/base/models.py
@@ -3,21 +3,14 @@ import json
 from django.contrib.auth.models import User
 from django.db import models
 
-import caching.base
 from tower import ugettext_lazy as _lazy
 
 from fjord.base import forms
 from fjord.base.validators import EnhancedURLValidator
 
 
-class ModelBase(caching.base.CachingMixin, models.Model):
-    """Common base model for all models: Implements caching."""
-
-    objects = caching.base.CachingManager()
-    uncached = models.Manager()
-
-    class Meta:
-        abstract = True
+# Common base model for all fjord models
+ModelBase = models.Model
 
 
 class Profile(models.Model):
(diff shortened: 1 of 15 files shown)


fjord/base/models.py
```diff
@@ -3,21 +3,14 @@
...
-class ModelBase(caching.base.CachingMixin, models.Model):
-    """Common base model for all models: Implements caching."""
-
-    objects = caching.base.CachingManager()
-    uncached = models.Manager()
-
-    class Meta:
-        abstract = True
+# Common base model for all fjord models
+ModelBase = models.Model
```
rlr: I left this instead of changing imports everywhere. But I could go either way. I have no idea if we'll need a common base class in the future vs the builtin django models.Model.

willkg: I think this is fine.

willkg:
I skimmed it and it looks fine. I'll test it tomorrow.

Then I'll make sure to push this by itself on Monday so I can do a before/after with New Relic and get a feel for how well it fares.

willkg:
I think we need to update this to use the regular django memcache cache thing:

https://github.com/mozilla/fjord/blob/a51449f79e5c591c5b36c4da177cdd6db9dc1ef7/fjord/base/tests/test_views.py#L53

Also, the comment here should be nixed and this class should derive from the new ModelBase:

https://github.com/mozilla/fjord/blob/master/fjord/search/models.py#L6

I skulked through the docs and the vagrant provisioning scripts and I think that's everything.

Can you fix the above two minor things? Then we'll land this on Monday so we can do the before/after thing.

rlr:
rebased and fixed those two things :point_up_2:

rlr:
565e05d [bug 1093222] Stop using django-cache-machine.
0a47fdc [bug 1093222] Remove django-cache-machine from peep requirements.
53d8043 [bug 1093222] Removing django-cache-machine from vendor.

PR #434, opened 2015-01-08
Fix bug 1114335 - purge_data script should record what happened in journal

lgp171188:
Removed the verbose print statements and instead logged the result of purge_data script in the journal.

### DIFF
--- a/fjord/feedback/models.py
+++ b/fjord/feedback/models.py
@@ -17,6 +17,7 @@ from fjord.base.models import ModelBase, JSONObjectField, EnhancedURLField
 from fjord.base.utils import smart_truncate, instance_to_key, is_url
 from fjord.feedback.config import CODE_TO_COUNTRY, ANALYSIS_STOPWORDS, TRUNCATE_LENGTH
 from fjord.feedback.utils import compute_grams
+from fjord.journal.utils import j_info
 from fjord.search.index import (
     register_mapping_type,
     FjordMappingType,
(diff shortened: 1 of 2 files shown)


fjord/feedback/models.py
```diff
@@ -750,18 +751,14 @@ def purge_data(cutoff=None, verbose=False):
...
 
     # Second, ResponseContext.
     objs = ResponseContext.objects.filter(opinion__created__lte=cutoff)
     responses_to_update.update(objs.values_list('opinion_id', flat=True))
     count = objs.count()
     objs.delete()
-
-    if verbose:
-        print 'Purged %d feedback_responsecontext records' % count
+    msg += 'feedback_responsecontext: %d, ' % (count, )
```
lgp171188: Not the best way to construct a string but does with minimal changes to the existing code

fjord/feedback/tests/test_models.py
```diff
@@ -322,6 +324,8 @@ def test_purge(self):
         eq_(ResponseContext.objects.filter(
             opinion__created__gte=cutoff).count(),
             5)
+        eq_(pre_journal_count+1,
```
willkg: Nit: There should be a spaces between things. This should be `eq_(pre_journal_count + 1,`.

fjord/feedback/tests/test_models.py
```diff
@@ -322,6 +324,8 @@ def test_purge(self):
...
+            Record.objects.filter(action='purge_data').count())
```
willkg: We might want to explicitly test the msg for the journal entry.

Also, interestingly, this is missing a ResponseRemoteTroubleshootingInfo portion. I'll make a note to add that after this lands.

lgp171188:
Oops, didn't notice it. Wonder why flake8 didn't complain! I have made the correction now.

fjord/feedback/tests/test_models.py
lgp171188: Since this is just a test, we can get rid of the query that finds the count by using our knowledge that there will be just one journal entry created when the testcase is run. So it can be rewritten as

    eq_(1,
            Record.objects.filter(action='purge_data').count())

I will also test for the value of the msg.

lgp171188:
I have incorporated all your suggestions.

fjord/feedback/models.py
willkg: This is fine. Another way to do it is to have msgs be a `[]` and append to it then join everything with a `', '`.

willkg:
Looks good to me! r+

willkg:
6661094 [bug 1114335] Record what happened when purge_data script ran, in journal

PR #439, opened 2015-01-09
[bug 1119015] Fix API with ES 1.2.4

willkg:
We were passing in a tuple to values_dict() which would pass it to ES in
the fields key like this:

    'fields': [('id', 'description', ...)],

the problem being that we need to pass it as a list of fields not a list
of a tuple of fields. This fixes that.

r?

### DIFF
--- a/fjord/feedback/api_views.py
+++ b/fjord/feedback/api_views.py
@@ -88,7 +88,7 @@ class PublicFeedbackAPI(rest_framework.views.APIView):
         search = search.order_by('-created')
 
         # Explicitly include only publicly visible fields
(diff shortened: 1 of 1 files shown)


willkg:
Best way to test is install Elasticsearch 1.2.4 and then run the tests. Without this fix, all the API tests fail. With this fix, they all pass.

rlr:
This makes sense to me. I haven't downloaded ES 1.2.4 and all that but I'll do that soon to check sumo. r+

willkg:
Thank you!

de4621e [bug 1119015] Fix API with ES 1.2.4

PR #441, opened 2015-01-09
Fix the _persist_feedbackdev decorator

willkg:
This fixes the decorator so that things that trace through it get the
right function name.

r?

This will make New Relic tracing more helpful.


rlr:
r+!

willkg:
Landed in 9d1ad87 Fix the _persist_feedbackdev decorator

Thank you!

PR #443, opened 2015-01-09
[bug 1116838] Create waffle flag for thank you page

aokoye:


willkg:
This looks ok except that we do all waffle flag migrations in the base app:

http://fjord.readthedocs.org/en/latest/db.html#data-migrations-for-data-in-non-fjord-apps

aokoye:
The latest commit should do it. I though I had deleted the initial waffle flag migration in the commit prior but quickly realized that that wasn't the case.

willkg:
Can you squash these commits into one?

aokoye:
I just squashed the commits.

willkg:
Awesome! Thank you!

9921759 [bug 1116838] Create waffle flag for thank you page

PR #446, opened 2015-01-12
Document how to configure the timezone of the VM to match the host

lgp171188:
This helps avoid odd timestamps in the commit history when the host computer is on a different timezone. Please feel free to suggest rewording of the text if appropriate.


willkg:
Looks great! Thank you!

PR #450, opened 2015-01-15
[bug 1071044] Initial travis setup.

rlr:
If I did things right, this just might trigger a Travis build.

### DIFF
--- a/bin/travis/setup.sh
+++ b/bin/travis/setup.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+# pwd is the git repo.
+set -e
+
+echo "Creating settings/local.py"
+cp fjord/settings/travis.py-dist fjord/settings/local.py
+
+echo "Creating test database"
+mysql -e 'CREATE DATABASE fjord CHARACTER SET utf8 COLLATE utf8_unicode_ci;'
+
+echo "Updating product details"
+python manage.py update_product_details
+
+echo "Starting ElasticSearch"
+pushd elasticsearch-0.90.10
+  # This will daemonize
+  ./bin/elasticsearch
+popd
+
+echo "Doing static dance."
+./manage.py collectstatic --noinput > /dev/null
+./manage.py compress_assets > /dev/null
--- a/bin/travis/test.sh
+++ b/bin/travis/test.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+# pwd is the git repo.
+set -e
+
+python manage.py test \
+  --noinput --logging-clear-handlers \
+  --with-nicedots
+echo 'Booyahkasha!'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,26 @@
+sudo: no
+language: python
+python: "2.7"
+node: "0.10"
+
+env:
+  globals:
+    - PIP_DOWNLOAD_CACHE="pip_cache"
+  matrix:
+    - TEST_SUITE=django
+
+cache:
+  directories:
+    - pip_cache
+
+install: bin/travis/install.sh
+before_script: bin/travis/setup.sh
+script: bin/travis/dispatch.sh $TEST_SUITE
+
+notifications:
+    email: false
+    irc:
+        channels:
+            - "irc.mozilla.org#input"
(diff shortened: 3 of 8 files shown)


bin/travis/setup.sh
```diff
@@ -0,0 +1,50 @@
...
+        'TEST_COLLATION': 'utf8_general_ci',
+    },
+}
+CELERY_ALWAYS_EAGER = True
+CACHE_BACKEND = 'caching.backends.locmem://'
+ES_INDEX_PREFIX = 'input'
+ES_URLS = ['http://localhost:9200']
+SECRET_KEY = 'TravisShhhhhhhhh'
+DEBUG = True
+SETTINGS
```
willkg: Instead of doing this, we should have a `fjord/fjord/settings/travis.py` file with the settings in it then set `DJANGO_SETTINGS_MODULE=fjord.settings.travis` in the environment.

bin/travis/setup.sh
```diff
@@ -0,0 +1,50 @@
...
+
+echo "Creating test database"
+mysql -e 'create database fjord'
```
willkg: I think this line should be more like the one in the `vagrant_provision.sh` script which sets the character set and collate:

https://github.com/mozilla/fjord/blob/451bd6d7f1ab22aa27f9b999739406c952ebec60/bin/vagrant_provision.sh#L44

bin/travis/setup.sh
```diff
@@ -0,0 +1,50 @@
...
+python manage.py update_product_details
+
+echo "Starting ElasticSearch"
+pushd elasticsearch-0.90.10
+  # This will daemonize
+  ./bin/elasticsearch
+popd
+
+echo "Starting XVFB for Selenium tests."
+/usr/bin/Xvfb :99 -ac -screen 0 1280x1024x16 >/dev/null 2>/dev/null &
```
willkg: We don't have selenium tests, so we can nix this.

bin/travis/test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+# pwd is the git repo.
+set -e
+
+# For XVFB Selenium tests.
+export DISPLAY=:99.0
+
+python manage.py test \
+  --noinput --logging-clear-handlers
```
willkg: We should add nosenicedots to `requirements/dev.txt` and then add `--with-nicedots`. It produces much better output.

willkg:
Looks good initially. I'm excited about this!

rlr:
Feedbacked so far :point_up: 

Right now it's peep installing compiled.txt, requirements.txt and dev.txt. I don't know if that has precedence over vendor or vice-versa. Maybe I should just peep install compiled.txt and dev.txt in the default build?

bin/travis/setup.sh
```diff
@@ -0,0 +1,22 @@
...
+set -e
+
+echo "Creating settings/local.py"
+cp fjord/settings/travis.py-dist fjord/settings/local.py
+
+echo "Creating test database"
+mysql -e 'CREATE DATABASE fjord CHARACTER SET utf8 COLLATE utf8_unicode_ci;'
+
+echo "Updating product details"
+python manage.py update_product_details
```
rlr: Do fjord tests need this? I wasn't sure but don't think so?

bin/travis/test.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+# pwd is the git repo.
+set -e
+
+# For XVFB Selenium tests.
+export DISPLAY=:99.0
```
rlr: oops. need to remove

bin/travis/setup.sh
willkg: Pretty sure they do. We have an events api that pulls from the product details data.

Having said that, I could be wrong. Might be worth doing a travis test without this line to know for sure.

willkg:
This looks great. Can you squash these? Then let's land it and tell jenkins to bugger off!

rlr:
yay travis is happy!

rlr:
The bug for travis was bug 1079357 #fail :sob:

PR #451, opened 2015-01-16
[bug 1116848] Change thank you page view

aokoye:


fjord/feedback/views.py
```diff
@@ -45,8 +46,15 @@ def download_firefox(request, template):
...
-    return render(request, template)
+    user_opinion = Response.objects.get(id=request.session['opinion_id'])
+    length = len(user_opinion.description.split())
+
+    if waffle.flag_is_active(request, 'thankyou') and user_opinion.locale == 'en-US' and user_opinion.happy == False and length >= 7:
+        template = 'feedback/thanks.html'
+        return render(request, template)
+    else:
+        template = 'feedback/thanks.html'
+        return render(request, template)
```
willkg: Lines 54 and 57 are identical and use a variable set in lines 53 and 56. So it's better to do it like this:

```
if blah blah blah:
    template = 'feedback/thanks.html'
else:
    template = 'feedback/thanks.html'
return render(request, template)
```

fjord/feedback/views.py
```diff
@@ -45,8 +46,15 @@ def download_firefox(request, template):
 
 
 def thanks(request):
-    template = 'feedback/thanks.html'
-    return render(request, template)
+    user_opinion = Response.objects.get(id=request.session['opinion_id'])
+    length = len(user_opinion.description.split())
+
+    if waffle.flag_is_active(request, 'thankyou') and user_opinion.locale == 'en-US' and user_opinion.happy == False and length >= 7:
+        template = 'feedback/thanks.html'
```
willkg: I was thinking we'd copy the `feedback/thanks.html` template to `feedback/thanks_sad.html`. That way we can write a series of tests to make sure this works correctly using `assertTemplateUsed`. Otherwise we have no way of testing that this code is working correctly.

fjord/feedback/views.py
```diff
@@ -45,8 +46,15 @@ def download_firefox(request, template):
 
 
 def thanks(request):
-    template = 'feedback/thanks.html'
-    return render(request, template)
+    user_opinion = Response.objects.get(id=request.session['opinion_id'])
+    length = len(user_opinion.description.split())
+
+    if waffle.flag_is_active(request, 'thankyou') and user_opinion.locale == 'en-US' and user_opinion.happy == False and length >= 7:
```
willkg: Nit: This line is too long. We have a 79 character limit for lines.

fjord/feedback/views.py
```diff
@@ -45,8 +46,15 @@ def download_firefox(request, template):
 
 
 def thanks(request):
-    template = 'feedback/thanks.html'
-    return render(request, template)
+    user_opinion = Response.objects.get(id=request.session['opinion_id'])
+    length = len(user_opinion.description.split())
```
willkg: Instead of calling this "length", I'd use something more descriptive like "word_count".

fjord/feedback/views.py
```diff
@@ -45,8 +46,15 @@ def download_firefox(request, template):
 
 
 def thanks(request):
-    template = 'feedback/thanks.html'
-    return render(request, template)
+    user_opinion = Response.objects.get(id=request.session['opinion_id'])
```
willkg: If there's no "opinion_id" in the session, then this will throw an error. I think it's worth unpacking this line into a couple of lines so that we can do proper error checking/handling.

fjord/feedback/views.py
```diff
@@ -22,6 +22,7 @@
 )
 from fjord.feedback import config
 from fjord.feedback import models
+from fjord.feedback.models import Response
 from fjord.feedback.forms import ResponseForm
```
willkg: Nit: We try to order imports alphabetically, so "from fjord.feedback.models..." should come after "from fjord.feedback.forms..."

fjord/feedback/views.py
```diff
@@ -45,7 +46,17 @@ def download_firefox(request, template):
 
 
 def thanks(request):
-    template = 'feedback/thanks.html'
+    opinion_id = request.session['opinion_id']
```
willkg: If opinion_id isn't in the session, I think this kicks up a `KeyError`. We should handle that scenario.

fjord/feedback/views.py
```diff
@@ -45,7 +46,17 @@ def download_firefox(request, template):
...
+    user_opinion = Response.objects.get(id=opinion_id)
```
willkg: Calling `.get()` can raise `MultipleObjectsReturned` (though with an id, that's super unlikely--our db would have to be totally fux0rd) and `DoesNotExist` (which is possible). We should handle the `DoesNotExist` scenario.

https://docs.djangoproject.com/en/1.6/ref/models/querysets/#get

fjord/feedback/views.py
```diff
@@ -45,7 +46,17 @@ def download_firefox(request, template):
...
+    word_count = len(user_opinion.description.split())
+
+    if (waffle.flag_is_active(request, 'thankyou') and
+        user_opinion.locale == 'en-US' and
+        user_opinion.happy == False and
+        word_count >= 7):
+        template = 'feedback/thanks_sad.html'
```
willkg: The if-stmt isn't readable since it's broken up into lines that are indented at the same place as line 57 which is the if-exp. There are a couple of ways to deal with this. You can either add an extra paren:

```
if ((blah and
     blah and
     blah):
    if-exp here
```

or you can indent an additional 4 spaces:

```
if (blah and
        blah and
        blah):
    if-exp
```

Either one is more readable.

willkg:
We also need to write a test for this to make sure it's triggering in the appropriate situations.

aokoye:
The latest commit has a test for the code.

fjord/feedback/tests/test_views.py
```diff
@@ -11,6 +11,7 @@
 from fjord.feedback.tests import ProductFactory
 
 
+
```
willkg: Nit: We don't need to add this extra line.

fjord/feedback/tests/test_views.py
```diff
@@ -102,6 +103,23 @@ def test_valid_sad(self):
         eq_(u'Firefox', feedback.product)
         eq_(u'14.0.1', feedback.version)
 
+    @with_waffle('thankyou', True)
+    def test_thank_you(self):
```
willkg: We can probably think of a better name for this method. Maybe something like "test_use_thank_you_sad" or something like that?

aokoye:
I took out that extra line and changed the name of the test.

fjord/feedback/tests/test_views.py
```diff
@@ -102,6 +103,23 @@ def test_valid_sad(self):
...
+        """Thanks_sad template is used when appropriate"""
+        url = reverse('feedback', args=(u'firefox',))
+        r = self.client.post(url, {
+            'happy': 0,
+            'description': u"I want to know why Firefox doesn't make me sandwiches!",
```
willkg: The sad thank you page requires a few things for it to get shown:

1. locale is en-US
2. # words > x
3. sad feedback

This test probably uses en-US, but it's not explicitly stated as such so if the test harness changes, this might break. Further, if the code this is testing isn't right and the test harness changes, we might not ever find out because this is not explicitly testing the en-US locale.

That should change.

Further, we need to make sure that changing any of those things returns the "normal" Thank You page.

fjord/feedback/tests/test_views.py
```diff
@@ -102,6 +103,23 @@ def test_valid_sad(self):
...
+        }, follow=True)
+
+        feedback = models.Response.objects.latest(field_name='id')
+        eq_(u"I want to know why Firefox doesn't make me sandwiches!", feedback.description)
+        eq_(u'', feedback.url)
+        eq_(False, feedback.happy)
+        eq_(u'en-US', feedback.locale)
+        eq_(r.status_code, 200)
+        self.assertTemplateUsed(r, 'feedback/thanks_sad.html')
```
willkg: This test is interested in the status code and the template used. Verifying the other things probably isn't important.

willkg:
@aokoye How're you doing with this?

aokoye:
I wrote a second test that makes sure that 'description's under 7 words doesn't redirect to the new thank you template.

fjord/feedback/tests/test_views.py
```diff
@@ -102,6 +102,37 @@ def test_valid_sad(self):
         eq_(u'Firefox', feedback.product)
         eq_(u'14.0.1', feedback.version)
 
+    @with_waffle('thankyou', True)
+    def test_use_thank_you_sad(self):
+        """Thanks_sad template is used when appropriate"""
+        url = reverse('feedback', args=(u'firefox',))
```
willkg: I just updated the reverse function:

https://github.com/mozilla/fjord/blob/d3fe0fcb975f7a81209b5e7d05d7f87e743f1a83/fjord/base/urlresolvers.py#L25

That change allows you to provide a `locale` argument to explicitly set the locale to use in the url. This line should explicitly set the locale to `en-US`.

fjord/feedback/tests/test_views.py
```diff
@@ -102,6 +102,37 @@ def test_valid_sad(self):
...
+        eq_(u'', feedback.url)
+        eq_(False, feedback.happy)
+        eq_(u'en-US', feedback.locale)
+        eq_(r.status_code, 200)
+        self.assertTemplateUsed(r, 'feedback/thanks_sad.html')
+
+    @with_waffle('thankyou', True)
+    def test_use_default_thank_you(self):
+        """Default thank you page is used when descirption is under 7 words"""
+        url = reverse('feedback', args=(u'firefox',))
```
willkg: Similar to the above, this line should explicitly set the locale to `en-US`.

fjord/feedback/tests/test_views.py
```diff
@@ -102,6 +102,37 @@ def test_valid_sad(self):
...
+        r = self.client.post(url, {
+            'happy': 0,
+            'description': u"Why doesn't it load?",
+        }, follow=True)
+
+        feedback = models.Response.objects.latest(field_name='id')
+        eq_(u'', feedback.url)
```
willkg: Why test the `feedback.url` here?

aokoye:
I had to fix a merge conflict but otherwise I edited the two previous tests and added one test to to see that the correct template gets if the locale is 'es' as opposed to 'en-US'.

aokoye:
The lasts of the tests have been written.

fjord/feedback/tests/test_views.py
```diff
@@ -102,6 +102,71 @@ def test_valid_sad(self):
...
+                'description': u'Tell me why Firefox is not making lunch.',
+            }, follow=True)
+
+            self.assertRedirects(r, reverse('thanks'))
+            feedback = models.Response.objects.latest(field_name='id')
+            eq_(u'es', feedback.locale)
+            self.assertTemplateUsed(r, 'feedback/thanks.html')
+
+        finally:
+            r = self.client.get('/en-US/feedback/')
```
willkg: This test is perfect!

fjord/feedback/tests/test_views.py
```diff
@@ -102,6 +102,71 @@ def test_valid_sad(self):
...
+
+        feedback = models.Response.objects.latest(field_name='id')
+        eq_(u'', feedback.url)
+        eq_(r.status_code, 200)
+        self.assertTemplateUsed(r, 'feedback/thanks.html')    
+
+    @with_waffle('thankyou', True)
+    def test_use_default_thank_you_happy(self):
+        """Default thank you page is used when the feedback is happy."""
+        url = reverse('feedback', args=(u'firefox',))
```
willkg: You should specify en-US here.

fjord/feedback/tests/test_views.py
```diff
@@ -102,6 +102,71 @@ def test_valid_sad(self):
...
+        r = self.client.post(url, {
+            'happy': 1,
+            'description': u"Firefox is the best browser I've ever used!",
+        }, follow=True)
+
+        feedback = models.Response.objects.latest(field_name='id')
+        eq_(u'', feedback.url)
```
willkg: Why test the feedback.url here?

fjord/feedback/tests/test_views.py
```diff
@@ -102,6 +102,71 @@ def test_valid_sad(self):
...
+        eq_(False, feedback.happy)
+        eq_(u'en-US', feedback.locale)
+        eq_(r.status_code, 200)
+        self.assertTemplateUsed(r, 'feedback/thanks_sad.html')
+
+    @with_waffle('thankyou', True)
+    def test_use_default_thank_you_words(self):
+        """Default thank you page is used when descirption is under 7 words"""
+        url = reverse('feedback', args=(u'firefox',))
```
willkg: This should specify en-US here.

fjord/feedback/tests/test_views.py
willkg: This line should specify en-US.

aokoye:
I made the changes you requested.

fjord/feedback/tests/test_views.py
willkg: Still missing the en-US on this line.

willkg:
Missing one last en-US. Other than that, this looks good to go. If you could add that and then squash the commits, then we'll merge it.

aokoye:
I added the locale to the first test and squashed the commits

willkg:
Awesome!

willkg:
3b7811f [bug 1116848] Change thank you page view

PR #452, opened 2015-01-16
l10n_completion tweaks

willkg:
* changed the percent from an integer to a float
* added an "untranslated_words" stat

r?

### DIFF
--- a/bin/l10n_completion.py
+++ b/bin/l10n_completion.py
@@ -76,15 +76,16 @@ def get_completion_data_for_file(fn):
                 path = path[1]
             else:
                 path = 'vendor/' + path[2]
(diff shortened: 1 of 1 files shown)


willkg:
@rlr, @mythmon I think we want to do this with kitsune, too. Otherwise the l10n stats don't move much. The one minor issue is that if the dashboard code is expecting an int, it'll be a float now and it will BREAK!

Someone asked me to add an "untranslated_words" stat so that they would know roughly how much work they have to do. Having said that, I think they asked me to do it in dennis and not postatus. But I think it's worth having in postatus.

rlr:
LGTM r+

willkg:
47b5584 l10n_completion tweaks

Thank you!

PR #455, opened 2015-01-22
[bug 1124497] table rename -- DO NOT MERGE

willkg:
Renames the ResponseTroubleshootingInfo table to ResponsePI.

r?

### DIFF
--- a/fjord/feedback/models.py
+++ b/fjord/feedback/models.py
@@ -586,6 +586,15 @@ class ResponseTroubleshootingInfo(ModelBase):
         return unicode(self.id)
 
 
+class ResponsePI(ModelBase):
+    """Holds remote-troubleshooting and other product data."""
+    opinion = models.ForeignKey(Response)
+    data = JSONObjectField()
+
+    def __unicode__(self):
+        return unicode(self.id)
+
+
 class NoNullsCharField(serializers.CharField):
     """Further restricts CharField so it doesn't accept nulls
 
--- a/fjord/feedback/models.py
+++ b/fjord/feedback/models.py
@@ -304,7 +304,7 @@ class Response(ModelBase):
 
     @property
     def has_browserdata(self):
-        if self.responsetroubleshootinginfo_set.exists():
+        if self.responsepi_set.exists():
             return True
         return False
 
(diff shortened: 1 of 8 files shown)


willkg:
Each commit should work on its own, migrations should work and the tests should pass.

Other than the fact that this is split into three commits, it should be pretty straightforward.

willkg:
Marking this as DO NOT MERGE because we're going to rebase this in so we have linear history to make it possible to deploy/undeploy individual commits.

willkg:
I did some testing just now with last night's export going forward and backward through the migrations and commits. I bumped into a couple of problems which I fixed and rebased in.

mythmon:
Protip: If you push the commits one at a time, Travis will run for all of them. When you push them all at once, it only runs for the last one.

willkg:
It's really hard to push commits to a PR one at a time. Definitely not something I'm going to do--I'd rather hand test things.

mythmon:
:+1: Makes sense. Testing locally also works very well. I just wanted to share some Travis foo :)

willkg:
^^^ That switches it to use `bulk_create`. I restricted it to a batch size of 500 figuring there are like 5000 items in the table and maybe 500 is a good number. The default is "do it all in one batch" and that seemed like a bad idea.

I had to rebase it in. Plus I rebased it against master tip. The new change are in the middle commit.

Does that look ok?

mythmon:
That uses a lot of DB queries. I think it will work though. I *think* the below might do something similar, but without as many queries. It might be slower to process though, since it is doing it all in memory. If you want to stop messing with it and go with what you have, that works for me. I'm happy either way.

PS: I haven't run any of the code you wrote, or this code below. Yours at least passed Travis, so that's a point in its favor.

```python
rtis = set(ResponseTroubleshootingInfo.objects.values_list('opinion', flat=True))
rpis = set(ResponsePI.objects.values_list('opinion', flat=True))
opinions_to_move = rtis - rpis
to_move = ResponseTroubleshootingInfo.objects.filter(opinion__in=opinions_to_move)
new_items = [ResponsePI(id=item.id, opinion=item.opinion, data=item.data)
             for item in to_move]
RepsonsePI.objects.bulk_create(new_items, batch_size=500)
```

willkg:
I'm not sure the "opinions_to_move = rtis - rpis" works unless the `__hash__` for equivalent ResponsePI and ResponseTroubleshootingInfo objects are the same. However, that's an interesting way to do it. I might do future data migrations that way.

In this case, there are only like 5000 items in the table so I'm not sure it's worth the additional work at this stage.

mythmon:
I assumed that the opinions were strings which do have a valid `__hash__`. Not important.

I don't have the ability to test it on a db dump though. That being said, I think it has all the properties you wanted, so I'd say :ship:.

willkg:
Oh, interesting. "opinion" is a ForeignKey to the Response table. I'm not sure if those lines end up as a set of Response ids or a set of Response objects. In either case, that'd work fine--you're right.

willkg:
Ok. Seems like this is "good enough", so I'm merging it now.

willkg:
Bah. I can't believe I just did that. (sigh)

willkg:
899c7a0 [bug 1124497] Delete ResponseTroubleshootingInfo
61a57e2 [bug 1124497] Move ResponseTroubleshootingInfo to ResponsePI
7398332 [bug 1124497] Add ResponsePI table and migration

PR #457, opened 2015-01-23
Tweak reverse to allow you to override the locale

willkg:
This makes it easier to use reverse in tests that involve non-en-US locales.
Before, we'd have to hard-code the url because reverse would do all this
stuff which would make it difficult to specify non-en-US locales. Now
we can do:

    url = reverse('some-thing', locale='es')

The changes were based on what Kitsune has.

r?

### DIFF
--- a/fjord/base/urlresolvers.py
+++ b/fjord/base/urlresolvers.py
@@ -1,6 +1,7 @@
 from threading import local
 
 from django.conf import settings
+from django.core.handlers.wsgi import WSGIRequest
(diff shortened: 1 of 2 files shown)


mythmon:
I read through this, and some of the surrounding code. This looks right, and matches what kitsune has. r+

That being said, this is the first time I've read how exactly we do locale selection, and the various bits and bobs in this file. I think a shower now.

willkg:
16250bf Tweak reverse to allow you to override the locale

Thank you!

PR #458, opened 2015-01-27
[bug 1054001] upgrade to django-browserid v0.11

rlr:
* There are changes to how redirects are done. Instead of a 302, it's a field in the JSON response.
* I had to register the helpers manually because django_browserid is failing to do it. @Osmose says he has seen it before or something similar and is just as puzzled.
* There is a new js file to include.

To test manually, I logged in with an existing user and a new user. WFM.

That's it. r? (CC: @Osmose can you eyeball this to see if it makes sense?)

### DIFF
--- a/fjord/base/tests/test_auth.py
+++ b/fjord/base/tests/test_auth.py
@@ -1,3 +1,5 @@
+import json
+
 from django.test.client import RequestFactory
 
 from nose.tools import eq_
@@ -20,17 +22,18 @@ class TestAuth(BaseTestCase):
         # FIXME - this can go away post django-browserid 0.9
         new_user.backend = 'django_browserid.auth.BrowserIDBackend'
 
-        get_request = RequestFactory().get(reverse('dashboard'))
-        get_request.user = new_user
-        get_request.session = self.client.session
+        post_request = RequestFactory().post(reverse('browserid.login'))
+        post_request.user = new_user
+        post_request.session = self.client.session
 
         fv = FjordVerify()
         fv.user = new_user
-        fv.request = get_request
+        fv.request = post_request
 
         resp = fv.login_success()
-        eq_(302, resp.status_code)
-        eq_(resp.get('location'), reverse('new-user-view'))
+        eq_(200, resp.status_code)
+        body = json.loads(resp.content)
+        eq_(body['redirect'], reverse('new-user-view'))
 
     def test_existing_user(self):
         """Tests that existing users get redirected to right place"""
@@ -44,27 +47,31 @@ class TestAuth(BaseTestCase):
         new_user.backend = 'django_browserid.auth.BrowserIDBackend'
 
         # First, do it RAW!
-        get_request = RequestFactory().get(reverse('dashboard'))
-        get_request.user = new_user
-        get_request.session = self.client.session
+        post_request = RequestFactory().post(reverse('browserid.login'))
+        post_request.user = new_user
+        post_request.session = self.client.session
 
         fv = FjordVerify()
         fv.user = new_user
-        fv.request = get_request
+        fv.request = post_request
 
         resp = fv.login_success()
-        eq_(302, resp.status_code)
-        eq_(resp.get('location'), '/')
+        eq_(200, resp.status_code)
+        body = json.loads(resp.content)
+        eq_(body['redirect'], '/')
 
         # Now do it with next!
-        get_request = RequestFactory().get(reverse('dashboard') + '?next=/foo')
-        get_request.user = new_user
-        get_request.session = self.client.session
+        post_request = RequestFactory().post(
+            reverse('browserid.login'),
+            {'next': '/foo'})
+        post_request.user = new_user
+        post_request.session = self.client.session
 
         fv = FjordVerify()
         fv.user = new_user
-        fv.request = get_request
+        fv.request = post_request
 
         resp = fv.login_success()
-        eq_(302, resp.status_code)
-        eq_(resp.get('location'), '/foo')
+        eq_(200, resp.status_code)
+        body = json.loads(resp.content)
+        eq_(body['redirect'], '/foo')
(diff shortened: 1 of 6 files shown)


Osmose:
Skimmed, makes sense and matches up with the expected upgrade path. Now that you've got a reproducible instance of the helper error I'll see what I can do about tracking the helper issue down.

requirements/requirements.txt
```diff
@@ -78,9 +78,9 @@ https://github.com/celery/billiard/archive/5652fe69e0a0623cea8f9f0c2810e001d4ad5
 # sha256: sSXkx6LhIZfZHt_5TesMpf_19yRaU11j8jH6P4mLWYw
 https://github.com/jsocol/bleach/archive/90a79d24fcce8f6404d9b636936e7886558495fb.tar.gz#egg=bleach
 
-# django-browserid: master~19
-# sha256: ENf2_W8JACouw1VJ96Nxl3ZX7Wox3ySOo4V0WocTXBc
-https://github.com/mozilla/django-browserid/archive/79ba1d01e742dfe95c08bb7517b9353e6c1d0050.tar.gz#egg=django-browserid
+# django-browserid: tags/v0.11
```
willkg: @osmose just released 0.11.1. We might as well pick that up.

rlr:
:point_up: went to v0.11.1

fjord/base/tests/test_auth.py
```diff
@@ -20,17 +22,18 @@ def test_new_user(self):
         # FIXME - this can go away post django-browserid 0.9
         new_user.backend = 'django_browserid.auth.BrowserIDBackend'
 
-        get_request = RequestFactory().get(reverse('dashboard'))
-        get_request.user = new_user
-        get_request.session = self.client.session
+        post_request = RequestFactory().post(reverse('dashboard'))
+        post_request.user = new_user
+        post_request.session = self.client.session
```
willkg: I don't understand this. Why is this posting to the dashboard?

rlr: Oops. I thought I understood these tests but I don't really.

rlr: I guess I don't test how getting the dashboard is testing auth?

fjord/base/tests/test_auth.py
```diff
@@ -44,27 +47,29 @@ def test_existing_user(self):
...
-        eq_(resp.get('location'), '/')
+        eq_(200, resp.status_code)
+        body = json.loads(resp.content)
+        eq_(body['redirect'], '/')
 
         # Now do it with next!
-        get_request = RequestFactory().get(reverse('dashboard') + '?next=/foo')
-        get_request.user = new_user
-        get_request.session = self.client.session
+        post_request = RequestFactory().post(reverse('dashboard'), {'next': '/foo'})
```
rlr: Oh, yeah. I changed it because `Verify` looks for `next` as a POST parameter. So this doesn't work the way it was originally tested.

rlr: See https://github.com/mozilla/django-browserid/blob/master/django_browserid/views.py#L38

So I can just kill this test. Not sure if it is needed. I don't know what's the use case of having `?next=...` on the dashboard page?

willkg: What happens when an existing user logs in? Do they end up in the right place?

fjord/base/tests/test_auth.py
willkg: Getting the dashboard didn't test the auth... it returned a get request which we then filled in and passed to `FjordVerify()` which did the authentication.

I'm not sure why we used the dashboard for that. That seems puzzling. When you click on "Log In", I'm pretty sure it authenticates against browserid urls.

So, I think this test was weird before and I think the changes continue the weirdness. Seems like we should fix this test.

This test covers what happens when people who don't have an account on Fjord log in with Persona. Since they're new to Fjord, they need to get passed to the `new-user-view`.

rlr:
OK. I think this is what is actually happening when you login ^^

fjord/base/tests/test_auth.py
```diff
@@ -20,17 +22,18 @@ def test_new_user(self):
         # FIXME - this can go away post django-browserid 0.9
         new_user.backend = 'django_browserid.auth.BrowserIDBackend'
```
willkg: Given that we're fiddling in here, can you see if this FIXME is right and that we can nix that line?

fjord/base/tests/test_auth.py
```diff
@@ -44,27 +47,31 @@ def test_existing_user(self):
...
         fv.user = new_user
-        fv.request = get_request
+        fv.request = post_request
 
         resp = fv.login_success()
-        eq_(302, resp.status_code)
-        eq_(resp.get('location'), '/foo')
+        eq_(200, resp.status_code)
+        body = json.loads(resp.content)
+        eq_(body['redirect'], '/foo')
```
willkg: This is interesting. I wonder if we have code that actually uses next. I think I'll push that investigation off to another day, though.

fjord/base/tests/test_auth.py
rlr: If I remove that, the test fails. Should I just remove the FIXME instead?

willkg: How puzzling. Nah--let's just leave it there because it's a curious thing someone should look into some day.

willkg:
This looks good. I'm going to pull it down and test it out locally for a bit, then if it's good, I'll merge it.

willkg:
Did some skulking around regarding where we use `next`.

https://github.com/mozilla/fjord/blob/fda4996da11460ed73c859b57514fb6252ab83fe/fjord/base/utils.py#L418

https://github.com/mozilla/fjord/blob/fda4996da11460ed73c859b57514fb6252ab83fe/fjord/base/views.py#L28

https://github.com/mozilla/fjord/blob/fda4996da11460ed73c859b57514fb6252ab83fe/fjord/base/browserid.py#L24

I think we need to sort that out. I think with the code changes in this PR, that `next` parameter will just get ignored, but we should either remove that `next` creation/handling code or fix it.

rlr:
OK, now I am passing next to the browserid button as the current page URL so that the user ends up on the same page after logging in.

willkg:
Works great! Want to squash them, then we'll merge and land.

rlr:
squashified :point_up:

willkg:
301e9e5 [bug 1054001] Upgrade to django-browserid v0.11.1

PR #460, opened 2015-01-28
[bug 1126030] Enable CORS for Heartbeat API.

rlr:
r?

### DIFF
--- a/fjord/heartbeat/views.py
+++ b/fjord/heartbeat/views.py
@@ -3,10 +3,18 @@ import rest_framework.response
 from statsd import statsd
 
 from .models import Answer, AnswerSerializer
+from fjord.base.utils import cors_enabled
 from fjord.journal.utils import j_error
 
 
 class HeartbeatV2API(rest_framework.views.APIView):
+
+    @classmethod
+    def as_view(cls, **initkwargs):
+        # Enable CORS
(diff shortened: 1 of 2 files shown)


fjord/heartbeat/views.py
```diff
@@ -3,10 +3,18 @@
...
 from fjord.journal.utils import j_error
 
 
 class HeartbeatV2API(rest_framework.views.APIView):
+
+    @classmethod
+    def as_view(cls, **initkwargs):
+        # Enable CORS
+        view = super(HeartbeatV2API, cls).as_view(**initkwargs)
+        return cors_enabled('*', methods=['POST'])(view)
```
willkg: Nice!

gregglind:
I think POST only is okay here, but fear it's missing something obvious.   We don't secretly allow PUT or want to support API access via GET right?  Dashboards go through some other mechanism?

willkg:
@gregglind: That API endpoint only does POST right now. Unless you need something different, it only needs CORS support for POST.

willkg:
This looks good to me. R+!

willkg:
ba498e3 [bug 1126030] Enable CORS for Heartbeat API.

willkg:
Thank you!

PR #461, opened 2015-01-28
Handle Windows 10 user agents in sniffer

willkg:
https://msdn.microsoft.com/en-us/library/ie/hh869301%28v=vs.85%29.aspx

That suggests that Windows 10.0 using browsers will have a user agent
with "Windows NT 10.0" in it rather than "Windows NT 6.4". Figure we can
handle both for now.

Quick r?


rlr:
r+!

willkg:
74dfc6d Handle Windows 10 user agents in sniffer

Thank you!

PR #468, opened 2015-01-30
[WIP] [bug 849077] Use django forms for dashboard controls

DESHRAJ:


fjord/analytics/forms.py
```diff
@@ -46,3 +46,13 @@ class Meta:
             'on_picker',
             'notes'
         ]
+
+class LeftDashboard(forms.Form):
```
mythmon: This is not a good name, because "Left" has nothing to do with what these controls are, and it isn't a dashboard. Maybe something like "DashboardFilterControls"?

fjord/analytics/views.py
```diff
@@ -191,19 +193,33 @@ def dashboard(request):
...
     # them to generate_dashboard_url.
-    search_happy = request.GET.get('happy', None)
-    search_platform = request.GET.get('platform', None)
-    search_locale = request.GET.get('locale', None)
-    search_product = request.GET.get('product', None)
-    search_version = request.GET.get('version', None)
+    if request.GET:
+        form = LeftDashboard(request.GET)
+        if form.is_valid():
+            search_happy = form.get('happy')
```
mythmon: You probably shouldn't be using these variables anymore, and instead access them all from form directly.

willkg:
This is a work-in-progress, right? If so, can you edit the summary to have "[WIP]" in it?

fjord/analytics/views.py
mythmon: Also, `form` doesn't have any method `.get()`, as you can see in [the Travis test failures](https://travis-ci.org/mozilla/fjord/builds/48937247#L4740).

DESHRAJ:
This bug requires huge changes in the code. After binding django form, it needs to replace the macro used in the template and write the new logic for displaying the fields which requires alot of changes to be done in analytics/dashboard.html and fjord/analytics/views.py . So use of django form will not bring the significant change and will consume alot of time and I think it should be closed as RESOLVED WONTFIX.

PR #470, opened 2015-01-30
[bug 1124379] Upgrade peep and un-goofify it

willkg:
* upgrade peep to 2.1.1
* rework how we handle the wheels problem so that instead of applying
  tweaks to peep directly which is hard to maintain and error-prone,
  we do our tweaks in a shell script and tell everyone to use that
* update the documentation, vagrant provision script and travis stuff
  to use the new peep.sh

r?

### DIFF
--- a/peep.sh
+++ b/peep.sh
@@ -1,0 +1,25 @@
+#!/bin/bash
+
+# Get pip version number
+PIPVER=`pip --version | awk '{print $2}'`
+ARGS=""
+
+echo "peep.sh: Using pip $PIPVER"
+
+# If we're using pip 1.5, 1.6 or 6.0, then toss on the --no-use-wheel
+# argument.
+#
+# Note: If we encounter other versions of pip here that require other
+# things, we can toss them in.
+case $PIPVER in
+    1\.5*|1\.6*|6\.*)
+        echo "peep.sh: Wheel-using pip detected, so passing --no-use-wheel."
+        ARGS="--no-use-wheel"
+        ;;
+esac
+
+# Execute peep with the command line args plus the additional args
+# if any.
+python ./bin/peep-2.1.1.py "$@" $ARGS
+
+echo "peep.sh: Done!"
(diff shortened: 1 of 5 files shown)


willkg:
I moved "peep" to "bin/peep-2.1.1.py" and in the process updated it. You can ignore all those changes.

If Travis fails, it's because I didn't fix the travis install.sh script correctly. I'll watch out for that.

Doing:

```
./peep.sh install -r requirements/requirements.txt
```

should do the right thing in regards to wheels now without people having to figure out which version of pip they have and adding the appropriate arguments.

peep.sh
```diff
@@ -0,0 +1,25 @@
...
+case $PIPVER in
+    1\.5*|1\.6*|6\.*)
+        echo "peep.sh: Wheel-using pip detected, so passing --no-use-wheel."
+        ARGS="--no-use-wheel"
+        ;;
+esac
+
+# Execute peep with the command line args plus the additional args
+# if any.
+python ./bin/peep-2.1.1.py $@ $ARGS
```
mythmon: You should put `$@` in double quotes, like `"$@"`. It's goofy, an edge case, and probably won't matter, unless some dingus has spaces in a path.

```bash
$ foo() { for i in $@; do echo $i; done }
$ foo "1 2" 3
1
2
3

$ foo() { for i in "$@"; do echo $i; done; }
$ foo "1 2" 3
1 2
3
```

willkg: Good point! That dingus will probably be me.

willkg:
^^^ (force push) That fixes the arguments issue in peep.sh.

mythmon:
I tested this by cloning a new copy of fjord, and running `./peep.sh -r requirements/compiled.txt -r requirements/dev.txt -r requirements/requirements.txt`. I did this with Pip 6.0.3 and pip 1.2 (to test a really old version without wheel support). It seems to have worked in both cases.

r+

willkg:
0bb2e56 [bug 1124379] Upgrade peep and un-goofify it

Thank you for looking at it!

PR #472, opened 2015-02-02
[bug 1111265] Fix browser data re: products and browsers

willkg:
Previously, we had the template code hardcoded to only collect browser
data if the product the user is leaving feedback for was "Firefox".
This prevents us from collecting browser data for "Firefox Dev Edition".

This changes the code so that the Product specifies whether to collect
browser data at all and which browser we'll collect browser data from.
For example, if the product has browser_data_browser=u'Android' and the
user is using Firefox for Android, then we'll add the browser data
collection bits to the template. Otherwise, we won't.

Also, I moved the "should we show the browser data section" bits from
the template to the view and added some better documentation.

r?

### DIFF
--- a/fjord/feedback/models.py
+++ b/fjord/feedback/models.py
@@ -85,6 +85,13 @@ class Product(ModelBase):
         max_length=20,
     )
 
+    # If this product should grab browser data, which browsers should
+    # it grab browser data for. Note, this value should match what
+    # we're inferring from the user agent.
+    browser_data_browser = models.CharField(
+        max_length=100, blank=True, default=u'',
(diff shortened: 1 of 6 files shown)


willkg:
Need to review the code and double-check the tests. Then if Travis is good, this is probably good.

When we deploy this, we'll need to go in and set the `browser_data_browser` fields.

willkg:
^^^ Sorry about that. I redid some bits moving some code into Product.

rlr:
Looks great to me and tests cover it nicely. r+

willkg:
deaa5eb [bug 1111265] Fix browser data re: products and browsers

Thank you!

PR #473, opened 2015-02-02
1111265 clean up

willkg:
I missed these when I added the new column.

Quick r?


rlr:
r+ yay for burying the dead code

willkg:
540b404 [bug 1111265] Add browser_data_browser to admin and forms

Thank you!

PR #477, opened 2015-02-04
[bug 1093341] Fix gengo warning emails

willkg:
We want an email warning that the Gengo account balance is low to be
sent out once per day rather than every time the sync stuff runs. This
adds a "daily activities" kind of thing to translation modules so that
we can run:

    ./manage.py translation_daily

and have that run all the things that should happen only once a day.

Additionally, we add a cron job so that this runs in the morning.

r?

### DIFF
--- a/bin/crontab/crontab.tpl
+++ b/bin/crontab/crontab.tpl
@@ -11,6 +11,9 @@ HOME=/tmp
 # to do an svn up.
 10 2 * * * root cd {{ source }} && (./bin/run_l10n_completion.sh {{ source }} {{ python }} > /dev/null)
 
+# Once a day at 3:00am run the translation daily activites.
+0 3 * * * {{ user }} cd {{ source }} && {{ python }} manage.py translation_daily -v 0 --traceback
+
 # Every hour, sync translations. This pulls and pushes to the various
 # translation systems.
 0 * * * * {{ user }} cd {{ source }} && {{ python }} manage.py translation_sync -v 0 --traceback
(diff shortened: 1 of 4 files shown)


willkg:
To review:

1. review the code changes
2. run the unit tests
3. run `./manage.py translation_sync`

Step 3 shouldn't do anything since you don't have gengo credentials. However, it should exist and not error out.

bin/crontab/crontab.tpl
```diff
@@ -11,6 +11,9 @@ HOME=/tmp
 # to do an svn up.
 10 2 * * * root cd {{ source }} && (./bin/run_l10n_completion.sh {{ source }} {{ python }} > /dev/null)
 
+# Once a day at 3:00am run the translation daily activites.
+0 * * * * {{ user }} cd {{ source }} && {{ python }} manage.py translation_daily -v 0 --traceback
```
rlr: You forgot the 3 in here?

rlr:
Other than that, this LGTM!

bin/crontab/crontab.tpl
willkg: OMG! Epic copy-paste fail. Bleh.

willkg:
^^^ That fixes the cron goof. Good catch!

rlr:
r+ :rocket:

willkg:
0fa7c28 [bug 1093341] Fix gengo warning emails

Thank you!

PR #480, opened 2015-02-06
[bug 1130009] Add pyflake and mccabe to requirements/dev.txt with hashes

lgp171188:
Add the missing dependencies along with their hashes in requirements/dev.txt so that peep will install it.


willkg:
9d009d7 [bug 1130009] Add pyflakes and mccabe to requirements/dev.txt with hashes

Thank you!

PR #481, opened 2015-02-06
[bug 1053384] Make the filters visible even when there are no results

lgp171188:
Set the count of selected filters to 0 before the actual count is calculated so that if the calculated count is zero, the filter is not excluded. Also fix the JavaScript code to handle 0 in the denominator when calculating the percentage.

Test for this fix has not been written since I have not used PyQuery before and have to figure out how test this.


willkg:
pyquery is just like using css selectors, so it's not too hard.

Having said that, the set up for this test is pretty finicky and dashboard edge cases like this aren't super critical, so having no test is fine.

willkg:
Landed in master in 39f3d25 [bug 1053384] Make the filters visible even when there are no results

Thank you!

PR #482, opened 2015-02-06
Prevent the heartbeat data table from overflowing outside containing div

lgp171188:
Add a 'overflow: scroll' style to the containing div so that the table
does not go outside it and instead there is a scrollbar to scroll to the
right and view all the columns and their values.


fjord/analytics/static/css/dashboard.less
```diff
@@ -31,6 +31,7 @@
 }
 
 .block.feedback {
+  overflow: scroll;
```
willkg: We don't want to do this here because it affects all the feedback blocks on the site including the front page dashboard.

It's better to add another css class with the `overflow:scroll` and add it to the specific div that's having difficulties.

lgp171188:
@willkg Created a separate class with the overflow property and applied it to the div in the particular template.

willkg:
Mmm... This doesn't really work. It means that the scrollbar for left-to-right is at the bottom of the page rather than at the bottom of the browser window which is a lot handier for someone who's trying to view all the data.

I'm not sure what other options we have. Maybe make the font size a little smaller? Maybe reduce the size of the th headers?

lgp171188:
Let me try out the options you have suggested and get back.

gregglind:
## PR comments

1.  All overflow options are gross :)  **perfered** I tend to support "overflow:hidden", and add a [+] element to turn it back into overflow:inherit.  This can be done client side.

## My goals of using this view are to:

1.  Debug specific rows.  For this case, I don't care about the 'table view' *at all*, other than to click on the link to the specific item to get to the item view.  This works *awesome*.
2.  To know if a 'record came in lately'.  To do this, I would sort by reverse `ts` and look if there is a row.  *this works awesome*
3.  Counts and flux.  For this, the table is not enough, and I would suggest a 'client side'  reporting tool that (using jquery or something) calculates (n last hour, n last day, % voted, mean scores) 'roughly'  at the client, on the displayed data.  This feels pretty worth pursuing.
4.  Weirdness detection in the dataset.  This is *such a big task* that it seems not worth pursuing.


## Specific suggestions

1.  If not done already, exporting the 'data' onto the window object might be helpful for dev, debugging, and experiments.  If I have `underscore`, `jQuery`, and the 'data' in the window, I can use devTools as my analysis env :)
2.  Do more at the client :)  Explore jQuery / tablify sore of things.
3.  adding some `class=fieldname` stuff might make styling, hiding by field a lot easier.

PR #488, opened 2015-02-11
[bug 1130469] Add a filter for hb test rows and distinguish them by color

lgp171188:
Add a filter for hiding test rows in hb survey answers table. Distinguish them by background color when not hidden.

### DIFF
--- a/fjord/analytics/static/css/dashboard.less
+++ b/fjord/analytics/static/css/dashboard.less
@@ -250,6 +250,9 @@ ul.bars {
     tr:nth-child(odd) {
       background-color: #eef;
     }
+    tr.test-answer {
+      background-color: #fdd;
+    }
   }
 }
 
--- a/fjord/analytics/analyzer_views.py
+++ b/fjord/analytics/analyzer_views.py
@@ -64,6 +64,18 @@ def hb_data(request, answerid=None):
         page = request.GET.get('page')
         answers = Answer.objects.order_by('-' + sortby)
         survey = request.GET.get('survey', survey)
+        showdata = request.GET.get('showdata', None)
+
+        if showdata:
+            if showdata == 'test':
+                answers = answers.filter(is_test=True)
+            elif showdata == 'notest':
+                answers = answers.exclude(is_test=True)
+            else:
+                showdata = 'all'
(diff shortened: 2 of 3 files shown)


fjord/analytics/static/css/dashboard.less
```diff
@@ -514,3 +514,7 @@ table.flagged {
     }
   }
 }
+
+.test-answer {
+   background-color: #fee !important;
```
willkg: We try to avoid `!important` because it involves funny things. It's better to make the css selector more specific.

In this case, you can define this rule in the `.summarytable` section since we only want it to apply to the summary table anyhow. Plus that might give you enough specificity that it overrides the `tr:nth-child` rules.

willkg:
After you fix the CSS issue, I'll test this out. But it looks good so far!

lgp171188:
Removed !important and used a more specific CSS selector to set the background color.

fjord/analytics/static/css/dashboard.less
```diff
@@ -514,3 +514,7 @@ table.flagged {
     }
   }
 }
+
+.summarytable tbody tr.test-answer {
+   background-color: #fee;
+}
```
willkg: LESS is hierarchical. This should be moved into the summarytable section of the file.

lgp171188:
Since I haven't used LESS before I had no idea :) Now I have fixed that.

willkg:
I looked at it and I think the `#fee` is too subtle. Can we change it to `#fdd`?

Also, I think the is_test filter needs to be able to show all three states of the filter: all, is_test=yes and is_test=no. The purpose of the filter is to filter out test items, but also to only show test items. These two modes make it easier for development work that Gregg is doing.

Given that I think it should be a drop-down and not a checkbox. Plus the filtering logic probably needs to change a little.

lgp171188:
@willkg I have changed the color to #fdd and made the filters into a dropdown as you have suggested.

willkg:
Looks great!

willkg:
be95d8e [bug 1130469] Add filter for hb test rows and distinguish them by color

Thank you!

PR #489, opened 2015-02-11
[bug 1130762] Alerts app, models and modelfactories

willkg:
This is the gist of the Alerts app.

r?

### DIFF
--- a/fjord/alerts/admin.py
+++ b/fjord/alerts/admin.py
@@ -1,0 +1,30 @@
+from django.contrib import admin
+
+from fjord.alerts.models import Alert, AlertFlavor, Link
+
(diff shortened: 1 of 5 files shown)


willkg:
This doesn't have any tests. Mostly it's a db schema.

When reviewing, just make sure it has the appropriate things for Django models. Otherwise this is sort of a no-op.

rlr:
New app LGTM. I ran the migration and it worked.

Bring on the Alerts! r+

willkg:
0fa0aab [bug 1130762] Alerts app, models and modelfactories

willkg:
Thank you!

PR #491, opened 2015-02-17
[bug 1133734] Fix waffle cookie thing

willkg:
The persist_feedbackdev decorator persists the querystring setting for
the feedbackdev flag into a cookie using a name that django-waffle will
key off of. That broke when we updated django-waffle since they moved
the value to a new module.

This fixes that to look at the correct module.

Quick r?


willkg:
Before this commit, if you went to:

```
/da/feedback/fxos?feedbackdev=1
```

it'd kick up an AttributeError. This fixes that.

rlr:
WFM r+

willkg:
f8cd339 [bug 1133734] Fix waffle cookie thing

Thank you!

PR #493, opened 2015-02-18
[bug 1103045] Align the create survey form like the bootstrap horizontal form

lgp171188:
Align the create survey form like the bootstrap horizontal form with the labels and the form controls aligned.


willkg:
The form looks like this on my computer:

http://bluesock.org/~willkg/images/link/1424353472.png

I have a couple of ideas:

First, is that maybe using `{{ form.as_ul() }}` will give conveniently structured HTML without us having to do it by hand. If you wrap that with a `div` of class `horizontal-form`, then that might work better.

Second, there's a `margin-bottom` on `p` from fjord.less. If you set `horizontal-form > p` `margin-bottom` to `20px`, that seems to fix some of the alignment issues.

PR #494, opened 2015-02-18
[bug 1130765] Implement Alerts API

willkg:
* add api_auth Token system for Django user agnostic API permission
  granting
* add alerts GET/POST API, docs and tests

r?

### DIFF
--- a/docs/alerts_api.rst
+++ b/docs/alerts_api.rst
@@ -1,0 +1,291 @@
+============
+ Alerts API
+============
+
+.. contents::
+   :local:
+
+This document covers Alerts v1 API.
+
+The Alerts v1 API is for posting and retrieving alert events. This
+data allows us to more easily keep track of trend changes as well as
+other events that can be related to user feedback.
+
+If you're interested in posting or retrieving Alerts data, speak to
+the User Advocacy group.
+
+
+Authentication token
+====================
+
+Posting to the API requires a valid enabled Token.
+
+To get a token, `write up a bug
+<https://bugzilla.mozilla.org/enter_bug.cgi?product=Input&rep_platform=all&op_sys=all>`_
+asking for a token. Make sure to state exactly what you'll be using
+the token for.
+
+To use a token, include the following HTTP header in your HTTP POST::
+
+    Authentication: token TOKEN
+
+where ``TOKEN`` is replaced by the 32-character token you received.
+
+.. Note::
+
+   Guard this token! If you reveal this token to the public, let us know
+   as soon as possible. We will disable the current one and issue a new
+   one.
+
+
+Alert flavors
+=============
+
+The alert flavor is the type of alert this is. It allows us to create
+a loose grouping of like alerts.
+
+For example, the "mfbt" alert flavor covers all alerts about mfbt.
+
+To get an alert flavor, `write up a bug
+<https://bugzilla.mozilla.org/enter_bug.cgi?product=Input&rep_platform=all&op_sys=all>`_
+asking for a flavor to be created. Please state exactly what you'll be
+using the flavor for. You should probably have a token first since
+without the token you can't emit any alerts.
+
+
+POSTing alerts
+==============
+
+:URL:            ``/api/v1/alerts/alert/``
+:Method:         HTTP POST
+:Payload format: JSON---make sure to have ``Content-type: application/json``
+                 header
+:Response:       JSON
+
+
+Example
+-------
+
+Using curl on the command line::
+
+    curl -v -XPOST 'https://input.mozilla.org/api/v1/alerts/alert/` \
+         -H 'Accept: application/json; indent=4' \
+         -H 'Content-Type: application/json' \
+         -H 'Authorization: token cd64de0e6c4c491f90fe1d362104c1e5' \
+         -d '
+    {
+        "severity": 5,
+        "summary": "mfbt now!",
+        "description": "it is 4:00pm on a friday and it is mfbt.",
+        "flavor": "mfbt",
+        "emitter_name": "mfbt-watcher",
+        "emitter_version": 1
+    }'
+
+
+Using Python requests:
+
+.. code-block:: python
+
+   import json
+
+   import requests
+
+   headers = {
+       'content-type': 'application/json',
+       'accept': 'application/json; indent=4',
+       'authorization': 'token cd64de0e6c4c491f90fe1d362104c1e5',
+   }
+   payload = {
+       'severity': 5,
+       'summary': 'mfbt now!',
+       'description': 'it is 4:00pm on a friday and it is mfbt.',
+       'flavor': 'mfbt',
+       'emitter_name': 'mfbt-watcher',
+       'emitter_version': 1,
+       'links': [
+           {'name': 'example', 'url': 'http://example.com'}
+       ]
+   }
+   resp = requests.post(
+       'https://input.mozilla.org/api/v1/alerts/alert/',
+       data=json.dumps(payload),
+       headers=headers
+   )
+
+   print resp.status_code
+   # 201
+   print resp.json()
+   # something like {u'detail': {u'id': 4}}
+
+
+Required fields
+---------------
+
+These fields are required and have no defaults. If you do not provide
+them, then you'll get back an HTTP 400 with a message stating you
+missed a required field.
+
++-------------------+--------+--------------------------------------------------------+
+|field              |type    |notes                                                   |
++===================+========+========================================================+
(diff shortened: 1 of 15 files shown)


willkg:
This is kind of a big PR. Sorry about that.

The only users for this will be Cheng and Robert. If there are bugs, it's not a big deal--we can fix them as we go along.

The things I'd love to get eyes on is the authorization/permission stuff and a review for "bad code".

fjord/alerts/admin.py
```diff
@@ -5,7 +5,7 @@
 
 class AlertFlavorAdmin(admin.ModelAdmin):
     list_display = (
-        'enabled', 'name', 'slug', 'description', 'more_info',
+        'name', 'slug', 'enabled', 'description', 'more_info',
```
willkg: Changing this means that "name" is the left-most column in the admin table and is the one that you click on for details. That's easier to use than "enabled".

fjord/urls.py
```diff
@@ -8,11 +8,12 @@
 from fjord.base.monkeypatches import patch
 patch()
 
-from django.contrib import admin
+from django.contrib import admin  # noqa
```
willkg: Had to add the `noqa` since we don't use it and flake8 got whiny.

docs/alerts_api.rst
```diff
@@ -0,0 +1,275 @@
...
+
++-------------------+-------+--------------------------------------------------------+
+|field              |type   |notes                                                   |
++===================+=======+========================================================+
+|severity           |integer|This is the severity of the alert. 0 = don't care. 10 = |
+|                   |       |call the president.                                     |
++-------------------+-------+--------------------------------------------------------+
+|summary            |string |Brief summary of what the alert is about.               |
++-------------------+-------+--------------------------------------------------------+
+|description        |string |Involved description fo all the details that help you   |
```
mythmon: Typo. "fo" -> "of"

docs/alerts_api.rst
```diff
@@ -0,0 +1,275 @@
...
+|                   |       |                                                        |
+|                   |       |You can make this a JSON encoded field if you have lots |
+|                   |       |of key/value pairs you want to include.                 |
++-------------------+-------+--------------------------------------------------------+
+|emitter_name       |string |Max length: 100.                                        |
+|                   |       |                                                        |
+|                   |       |The name of the emitter that created this alert. It     |
+|                   |       |could be a script name. It could be a sekret code-name  |
+|                   |       |for the emitter. It could be skynet. So long as we can  |
+|                   |       |distinguish it from other emitters, it's all good.      |
```
mythmon: I think you missed a good opportunity for a skynet joke here. Oh well.

mythmon:
The auth and permissions look good. I'm glad that turned out to be straight forward. I assume the only way to create new flavors and tokens (and link the two) is you doing it in the admin?

Is there a reason you made the API views yourself instead of using the model mixins like [here](https://github.com/mozilla/kitsune/blob/master/kitsune/notifications/api.py#L66-L80)? I think it would simplify those a lot, though the thing you do with Links might not be doable, which would be a reason to do what you did instead of using the mixins. That would require more research, I guess.

This seems solid, and I think it achieve the goals we talked about for auth. I didn't run the code or anything, I only looked at the diffs, but Travis liked it.

willkg:
I never got around to becoming enough of an expert in DRF to use the mixins effectively without a lot of struggling. I used them in richard (https://github.com/pyvideo/richard/blob/master/src/richard/videos/views.py#L316), but I find it a lot easier and maintainable to just go with subclassing APIView and doing my own thing. I have similar issues with Django CBVs. DRF 3 makes some API changes. It's possible that'll make it easier.

In this case, I'm not sure it helps a whole lot. Flavor comes from the url. Links are slightly complicated (though I've done similar stuff with richard).

I'm on the fence. I could rewrite it with generics. Do you think it'd be easier to augment and maintain?

Speaking of flavor, I'm not sure what I did is such a great idea. Maybe I should change it to `POST /api/v1/alerts/` and `GET /api/v1/alerts/<flavor>/`?

mythmon:
One thing that could be better in that Richard view would be moving the filtering to a `filter_backend`, like [in the SUMO Question's API](https://github.com/mozilla/kitsune/blob/master/kitsune/questions/api.py#L132-L211). The filter backends tend to be more declarative, which I find helps with keeping things simple. Also, I'm pretty sure `get_queryset` isn't meant to do any filtering anyways, you would do that in something like `filter_queryset`. This is partially because `get_queryset` is used as the base for detail views too, not just list views.

About the url structure: Does it make sense to get a list of alerts without filtering on flavor? Like `GET /api/v1/alerts/`? If so, I'd actually make the api `GET /api/v1/alerts/?flavor=<flavor>` instead. At least, that is what I've been doing on on SUMO's v2 api. The general pattern I've been using is `/api/2/<model>/[id]/[action]/[?filter=value]`. The id and action don't apply to all queries, so sometimes that is `/api/2/user/generate` (which is the generate action on the user "model") or `/api/2/question/1234/` (which is the detail view for the 1234 question). Filters are always query parameters, and don't really apply to detail views.

> NB: Above the term "model" doesn't map 1:1 with Django's models, mainly because the "user" API model is actually a mix of `django.contrib.auth.models.User` and `kitsune.users.models.Profile`.

As far as switching to generics... I'm not sure. It looks to me like a lot of what you are doing is error checking and input validation, which the generics are really good with. On the other hand, I think that it would be hard to convince DRF 2.x to do what you want with the links. It also isn't really a single model API, since you tie a lot of behavior into the Flavor class stuff (which makes the generics less useful). I'd say if you want a non-trivial generics view to learn with, this would be that. In the end, maintainability is about who is writing the code, so if this is easier for you to maintain, I'd say stick with it. I don't think what you have is bad. I like CBV and the DRF generics mainly because they tend to make smaller functions (though more of them) instead of big monolithic view functions.

willkg:
That makes sense. I think these view functions are small enough I'll just stick with what I've got. They're short and easy to reason about. I do think I'll rework the API urls. That'll reduce some of the complexity.

 I'll look into the richard stuff next time I'm fiddling with that API.

Thanks!

docs/alerts_api.rst
```diff
@@ -0,0 +1,275 @@
...
+
+Alert flavors
+=============
+
+The alert flavor is the type of alert this is. It allows us to create
+a loose grouping of like alerts.
+
+For example, the "mfbt" alert flavor covers all alerts about mfbt.
+
+To get a token, `write up a bug
```
rlr: `To get a token,` doesn't make sense here. I think it's a copy/pasta mistake?

docs/alerts_api.rst
willkg: I was super tired writing this and dealing with emacs table-mode. I'm totally game for anything you have in mind.

willkg:
I factored the comments in (and fixed the docs typos) and rewrote the API endpoints like this:

POST `/api/v1/alerts/flavor/<flavor>/addalert/`

> POST pulls the flavor from the url, but everything else from the HTTP request body. There are good tests for allt he bits.

GET `/api/v1/alerts/alert/`

> GET can take two querystring arguments: flavors and max. It verifies each flavor before doing anything. There are good tests for all the bits.

willkg:
^^^ That fixes the documentation examples.

willkg:
This is ready for another r? round.

mythmon:
It seems odd to use an action on the `/alerts/flavor` endpoint to create an alert, instead of using `POST /api/v1/alerts/alert {"flavor": "mfbt", ...}`. Flavor is a property of alerts, not the other way around, right?

willkg:
Mmm... Flavor is a property of alerts, yes, but there's a permission component here so I think the better way to talk about this is less "I'm creating an alert" and more "For this flavor, I'm adding an alert" where the operation is on the flavor's alerts, not alerts in general. If there wasn't an authorization component on the flavor, I'd probably just have it be alerts in general.

mythmon:
That doesn't seem like a good enough reason to have a surprising URL for creating alerts, or for having  asymmetry between creating and viewing alerts. Even the English translation ("for this flavor, I'm adding an alert") sounds awkward. I don't think it is any easier to code either way, but I do like consistent URL structures, and this doesn't feel consistent to me.

willkg:
^^^ Changed the api endpoint to what you suggested. IT WILL BE BETTER!

mythmon:
I've got nothing else to talk about. This looks good, well tested, and well documented. I say :boat:.

willkg:
e712384 [bug 1130765] Implement Alerts API

Thank you both for working through this with me!

PR #495, opened 2015-02-22
[bug 1130765] Add support for Fjord-Authorization header

willkg:
ZLB isn't passing along Authorization header. This adds support for
looking for a Fjord-Authorization header if Authorization isn't there.

Note: This is a stop-gap. As soon as we can ditch this, we should.

r?

### DIFF
--- a/docs/alerts_api.rst
+++ b/docs/alerts_api.rst
@@ -27,7 +27,7 @@ the token for.
 
 To use a token, include the following HTTP header in your HTTP POST::
 
-    Authorization: Token <TOKEN>
+    Fjord-Authorization: Token <TOKEN>
 
 where ``<TOKEN>`` is replaced by the token you received.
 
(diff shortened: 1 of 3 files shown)


mythmon:
I'd add a note to the docs like "This uses a non-standard authorization header instead of the normal one because of bug 1234.", and probably add that to the comment block explaining the new header too. The actual mechanics and the test look good, so r+wc.

PS: I was going to recommend that this should be X-Fjord-Authorization, since I thought `X-` was supposed to be used for custom headers. Turns out that is false, and in fact new things should not use `X-`, at least according to [RFC 6648](http://tools.ietf.org/html/rfc6648).

The more you know

willkg:
I thought about adding a note to the docs along those lines, but then decided that particular note doesn't really help a user using the API. It might be a "nice-to-know", but there's nothing the user can do about it now or in the future. I'm trying to keep the user docs small where possible.

One possible future is that this stays like this forever.

Another possible future is that when we switch to AWS, we try to nix this silly thing. If we can, then we'll deprecate the Fjord-Authorization header, update the docs, and notify emitter users. Then after emitters stop using the new header, we nix it.

In either case, I don't think the note helps and I'm concerned it adds more confusion for emitter writers.

mythmon:
That sounds reasonable.

I don't think Travis ran for this PR because, for whatever reason, GitHub can't automatically merge the pull request. As usual, I haven't tested this (I should probably set up a local Fjord instance... I hear Vagrant makes that easy). On the one hand, it may be a good idea to fix the merge problem so Travis can run, but on the other, the test seems straight forward, I'm sure you've run the tests locally, and these APIs aren't critical to the site, so I'm fine either way.

willkg:
Oh screwy--github created this pull request branch off of my testing branch and not master which is what it branched from. I'll see if I can fix that, but if not, I'll recreate the PR.

PR #497, opened 2015-02-23
[bug 909011] Handle amqp-specific indexing errors

willkg:
This tweaks the code for live indexing so that we "ignore" amqp-related
issues. This will reduce the HTTP 500 errors users see when leaving
feedback during a rabbitmq outage.

r?

### DIFF
--- a/fjord/search/tasks.py
+++ b/fjord/search/tasks.py
@@ -1,8 +1,10 @@
 import datetime
 import logging
 import sys
+import traceback
 
 from django.conf import settings
+from django.core.mail import mail_admins
 from django.db.models.signals import post_save, pre_delete
 
 from celery import task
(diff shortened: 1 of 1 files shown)


willkg:
There are no tests for this, but you can simulate it a bit like this:

1. set `CELERY_ALWAYS_EAGER` to `False`; make sure rabbitmq is not running
2. leave feedback--you should see a Thank you page; if you check the dashboard, the response won't show up because it never got indexed
3. make sure rabbitmq is running
4. leave feedback--you should see a Thank you page; if you check the dashboard, the response will show up

The only way to test this more programmatically is to mock amqp, but that gets me into mocking the guts of amqp with things I don't know much about and that seems to contrived to be wildly helpful. Having said that, I could be convinced that this is is more helpful than I'm thinking and that I should do it.

fjord/search/tasks.py
```diff
@@ -123,13 +125,51 @@ def _live_index_handler(sender, **kwargs):
...
+        # send an email. We can collect reasons and narrow this down
+        # at some point if that makes sense to do. If "amqp" is not in
+        # the frames, then it's some other kind of error that we want
+        # to show up, so we'll re-raise it. Sorry, user!
+        #
+        # In this way, users will stop seeing HTTP 500 errors during
+        # rabbitmq outages.
+        exc_type, exc_value, exc_tb = sys.exc_info()
+        frames = traceback.extract_tb(exc_tb)
+        for fn, ln, fun, text in frames:
```
rlr: so much fun!

willkg: fun in every frame!

rlr:
Works as advertised. r+ :shipit:

willkg:
1f3bde0 [bug 909011] Handle amqp-specific indexing errors

Thank you!

PR #500, opened 2015-02-26
1136809 multiple obs

willkg:
r?

### DIFF
--- a/docs/db.rst
+++ b/docs/db.rst
@@ -26,13 +26,18 @@ To create a new migration the automatic way:
 
    where ``<app>`` is the app name (base, feedback, analytics, ...).
 
-3. run the migration on your machine::
+3. add a module-level docstring to the new migration file specifying
+   what it's doing since we can't easily infer that from the code
+   because the code shows the new state and not the differences
+   between the old state and the new stage
+
+4. run the migration on your machine::
 
        ./manage.py migrate
 
-4. run the tests to make sure everything works
-5. add the new migration files to git
-6. commit
+5. run the tests to make sure everything works
+6. add the new migration files to git
+7. commit
 
 
 .. seealso::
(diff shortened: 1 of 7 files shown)


willkg:
Run the tests and eyeball the code.

Pretty sure the problem is a race condition. We're ramping up Heartbeat, so it's going to be more prevalent.

I'm pretty sure this is the best thing I can do where "best" is defined as "least amount of icky code". I'm game for other solutions, too.

Pretty sure the race condition will kick up HTTP 500 errors. I'm going to wait to fix those until I have more data. Plus I need to fix bug #1136840 so I have more data to work with.

willkg:
Also, ...

PULL 500!!!!!

fjord/alerts/migrations/0003_auto_20150225_1213.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
rlr: you can rename the migration files to something more descriptive if you want. It would make browsing the migrations folder easier but that's kind of a rare use case.

rlr:
Looks good. R+

willkg:
We had a long chat on IRC about some of the problems in this PR:

1. Data migrations end up with "auto" in the filename which is goofy since data migrations aren't auto-generated. @rlr pointed out that this is fixed in Django 1.8. For now, I'll fix the filename by hand.
2. Migration file names aren't indicative of what they're doing like they were in South. We talked about requiring better filenames, but that's a pain in the ass with the current `makemigrations` command. @rlr pointed out that Django 1.8 has a `--name` argument. For now, we'll just leave it as is and when we upgrade to 1.8, we'll change the docs to require `--name` be specified.
3. We should require that migration files have a module-level docstring explaining wtf they're doing. Otherwise this information gets lost in the sands of time.
4. When we change `help_text`, it generates a migration. While the migration doesn't actually do anything to the db, it is executing some SQL. For now, we're going to fold the `help_text` changes into the previous migration. See https://code.djangoproject.com/ticket/21498 and http://stackoverflow.com/a/26513612

I'll update the PR now.

willkg:
^^^ That nixes some migrations that were "no-ops" (with scare quotes), tweaks the docs regarding module-level doc-strings, adds some doc-strings and makes a valiant but failed attempt to solve the Traveling Salesman Problem using only one chopstick.

rlr:
that's better! :tada: r+

willkg:
Landed in master:

3ce45ec Update schema migration docs regarding module-level docstrings
2d3d05a [bug 1136809] Fix (person, survey, flow) uniqueness issues
473e900 Fix migration for fixing AlertFlavor.allowed_tokens
1721758 Update help_text in api_auth initial migration

Thank you for reviewing this! That helped a ton!

PR #501, opened 2015-02-26
[bug 1137430] Validate sortby values

willkg:
This requires the sortby querystring arg to be one of the specified
valid values. Otherwise we make it 'id'.

Quick r?


willkg:
There aren't any tests for this code, so we should just eyeball it.

Long term, this code isn't going to change much and it shouldn't be in Input forever. @lgp171188 is rewriting the hb survey views to use CBVs. I think we'll base future view work off of that code rather than this hb data view code.

So that leaves the short term where we have a bunch of minor issues. Figure it's good to fix the easy ones and we'll deal with the hard ones later.

rlr:
r+

willkg:
2a91627 [bug 1137430] Validate sortby values

Thank you!

PR #504, opened 2015-03-02
[bug 1103045] Add update survey form

lgp171188:
Add a form for updating surveys using the CBVs and the same logic and template used in the create survey form


willkg:
When you click on "update" link, it brings you to an update page with the list of surveys above. I think I'd just make that page just show the update survey form and not have the list of surveys above it. Further, it needs a "Cancel" button or a "Back to surveys" link or something like that to go back to the survey list page.

Other than that, this looks good to me!

lgp171188:
I have implemented the changes you have suggested. Please review and let me know if there is anything else to fix.

willkg:
Awesome! Thank you!

willkg:
Landed in 3668585 [bug 1103045] Add update survey form

PR #505, opened 2015-03-02
[bug 1048462] Update some requirements to pinned versions.

rlr:
This updates the requirements that already matched released versions
to use pypi pinned versions instead of github URLs.

There are still a dozen or so github URLs that need to be sorted out.

r?


willkg:
I created a new vagrant environment while I was working on some other things and then ran the tests with this and everything seems fine, so I think we're good!

willkg:
ac35deb [bug 1048462] Update some requirements to pinned versions.

PR #506, opened 2015-03-02
[bug 1137839] alerts timestamps

willkg:
This adds start_time and end_time to the alerts api.

There isn't any code that uses it. Pretty sure the tests cover most/all situations.

r?

### DIFF
--- a/fjord/alerts/models.py
+++ b/fjord/alerts/models.py
@@ -1,5 +1,9 @@
+from django.conf import settings
+from django.core.exceptions import ValidationError
 from django.db import models
 
+import pytz
+from rest_framework import fields
 from rest_framework import serializers
 
 from fjord.api_auth.models import Token
@@ -50,6 +54,12 @@ class Alert(ModelBase):
         help_text=u'Unique name for the emitter that created this')
     emitter_version = models.IntegerField(
         help_text=u'Integer version number for the emitter')
+    start_time = models.DateTimeField(
+        null=True, blank=True,
+        help_text=u'Timestamp for the beginning of this event')
+    end_time = models.DateTimeField(
+        null=True, blank=True,
+        help_text=u'Timestamp for the end of this event')
     created = models.DateTimeField(auto_now_add=True)
 
     def __unicode__(self):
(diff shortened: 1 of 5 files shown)


willkg:
@mythmon Does the `from_native` thing look good to you?

fjord/alerts/models.py
```diff
@@ -69,6 +79,28 @@ def __unicode__(self):
         return self.url
 
 
+class UTCDateTimeField(fields.DateTimeField):
+    """Like DateTimeField, except it always in UTC"""
```
mythmon: Grammar problem. You probably want `s/it/it's/`

willkg: Actually, I'm trying out docstrings out of East Orange, NJ.

"You down with UTC? YEAH YOU KNOW ME!"

fjord/alerts/models.py
```diff
@@ -69,6 +79,28 @@ def __unicode__(self):
...
+        result = super(UTCDateTimeField, self).from_native(value)
+        # We're using MySQL with settings.USE_TZ=False. So we need to
+        # peel off the timezone if we get sent one.
+        if result is not None and result.tzinfo is not None:
+            # First convert it to UTC
+            result = result.astimezone(pytz.UTC)
+
+            # Then drop the timezone
+            result = result.replace(tzinfo=None)
+        return result
```
mythmon: I'm not sure this is quite right. I think this will end up shifting dates by 8 hours every time (because 8 hours is Input's offset from UTC). I'm imagining this:

1. I decide that something happened, so I send call the alert API. Like a good user, I encode the timestamp in ISO8601, in the UTC time zone. Something like `2015-03-02T17:06:35Z`.
2. Line 92 converts it to a non-naive Python datetime object (in UTC, representing the right time). Something like `datetime.datetime(2015, 3, 2, 17, 6, 35, 0, tzinfo=<UTC>)`.
3. Line 97 converts it to UTC, which is a no-op.
4. Line 100 drops the time zone info. That will turn it from a non-naive datetime to a naive one with the same numbers, like `datetime.datetime(2015, 3, 2, 17, 6, 35, 0, tzinfo=None)`.
5. That goes in the DB.
6. Later, I want to know about that alert, so I ask the API. The timestamp is fetched as a naive datetime from the DB.
7. Line 87 converts the timestamp into a non-naive datetime object in PST without changing the numbers. Something like `datetime.datetime(2015, 3, 2, 17, 6, 35, 0, tzinfo=<DstTzInfo 'US/Pacific' PST-1 day, 16:00:00 STD>)`
8. Line 88 converts that to UTC, resulting in something like `datetime.datetime(2015, 3, 3, 1, 06, 35, 0, tzinfo=<UTC>)`. Importantly, this step changes the hour (and the day, but just because time wraps).
9. That datetime is serialized into a string to be sent out in JSON. It looks like `2015-03-03T01:06:35Z`. Not what we started with.

Does that logic sound correct? I think what you need to do is convert to `pytz.timezone(settings.TIME_ZONE)` on line 97, instead of UTC.

willkg: Oh, whoops--you're right!

I was thinking I'd store the data as UTC, but now that I'm reflecting on that and the fact that the timestamps in other tables is is `settings.TIME_ZONE` time and the `created` field will be in `settings.TIME_ZONE` time, I should also store `start_time` and `end_time` as `settings.TIME_ZONE` time.

Thank you for catching this. I'll rewrite it tomorrow.

willkg: Just to clarify "I was thinking I'd store the data as UTC..." was just what I was thinking about when I wrote this, but regardless I completely forgot to match `from_native` up with `to_native`. The disconnect between the two is the crux of the problem ehre.

willkg:
^^^ That fixes the typo and also fixes `to_native` and `from_native` to match.

@mythmon Does that look right? I also added a "roundtrip" test to make sure we POST an alert and then GET the same data back out.

mythmon:
The new version of the field looks right, and the round trip test looks good (if a bit wordy). A docstring needs fixed, and I think you could easily fix a bunch of the FIXMEs in the tests with `@override_settings`. Besides that, it looks good.

willkg:
Oh, you're right! I should totally do the `@override_settings` thing. I'm feeling really dumb today.

willkg:
Landed in master in 7a60497 [bug 1137839] Add start_time/end_time to alerts api

Thank you so much for the review work!

PR #509, opened 2015-03-04
[bug 1136840] Fix error handling for better debugging: WIP

willkg:
* creates DebuggableWSGIHandler which should send more useful error
  emails so I can debug API-related problems, plus in DEBUG=True
  mode, it'll return text/plain content rather than text/html content
  which is a lot easier to read in a terminal when debugging API-related
  problems
* created fjord/wsgi.py which is where the wsgi stuff *should* be--we'll
  have to get the servers fixed to use this instead of wsgi/playdoh.wsgi
  if this works
* for now, wsgi/playdoh.wsgi and fjord/wsgi.py should be identical--the
  former is used in production and the latter is used in development,
  but should be used in production

r?

### DIFF
--- a/fjord/wsgi_utils.py
+++ b/fjord/wsgi_utils.py
@@ -1,0 +1,87 @@
+from django.conf import settings
+
+
+class BetterDebugMixin(object):
+    """Provides better debugging data
+
+    Developing API endpoints and tired of wading through HTML for HTTP
+    500 errors?
+
+    Working on POST API debugging and not seeing the POST data show up
+    in the error logs/emails?
+
+    Then this mixin is for you!
+
+    It:
+
+    * spits out text rather than html when DEBUG = True (OMG! THANK
+      YOU!)
+    * adds a "HTTP_X_POSTBODY" META variable so you can see the raw post
+      data in error emails which is gross, but I couldn't figure out
+      a better way to do it
+
+    Usage:
+
+    Create a WSGIHandler subclass and bind that to ``application`` in
+    your wsgi file. For example::
+
+        import django
+        from django.core.handlers.wsgi import WSGIHandler
+
+        from fjord.wsgi_utils import BetterDebugMixin
+
+
+        class MyWSGIHandler(BetterDebugMixin, WSGIHandler):
+            pass
+
+
+        def get_debuggable_wsgi_application():
+            # This does the same thing as
+            # django.core.wsgi.get_wsgi_application except
+            # it returns a different WSGIHandler.
+            django.setup()
+            return MyWSGIHandler()
+
+
+        application = get_debuggable_wsgi_application()
+
+    """
+    def handle_uncaught_exception(self, request, resolver, exc_info):
+        if settings.DEBUG_PROPAGATE_EXCEPTIONS:
+            raise
+
+        # First, grab the raw POST body and put it somewhere that's
+        # guaranteed to show up in the error email.
+
+        # This should be "bytes" which is str type in Python 2.
+        #
+        # FIXME: This is probably broken with Python 3.
+        postbody = getattr(request, 'body', '')
+        try:
+            # For string-ish data, we truncate and decode/re-encode in
+            # utf-8.
+            postbody = postbody[:10000].decode('utf-8').encode('utf-8')
+        except (UnicodeDecodeError, UnicodeEncodeError):
+            # For binary, we say, 'BINARY CONTENT'
+            postbody = 'BINARY OR NON-UTF-8 CONTENT'
+
+        # The logger.error generates a record which can get handled by
+        # the AdminEmailHandler. Overriding all that machinery is
+        # daunting, so we're instead going to shove it in the META
+        # section which shows up when the machinery does a repr on
+        # WSGIRequest.
+        request.META['HTTP_X_POST_BODY'] = postbody
+
+        # Second, check the Accept header and if it's not text/html,
+        # pretend this is an AJAX request so that we get the output in
+        # text rather than html when DEBUG=True.
+        if settings.DEBUG:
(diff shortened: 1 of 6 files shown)


willkg:
Since I added `WSGI_APPLICATION` to settings, doing `./manage.py runserver` and `./manage.py test` will use `fjord/wsgi.py` to build the WSGI application. So running the tests will give us some confidence that the `wsgi.py` file is at least "mostly correct".

However `wsgi/playdoh.wsgi` which runs on stage/prod does some things that will mess up the contributor environment. I still need to figure out how to deal with that. Maybe I'll do something goofy like have `wsgi/playdoh.wsgi` do the `os.environ` stuff and then call out to `fjord/wsgi.py` to do the rest.

The tests don't kick up any exceptions, so there's nothing going through that new section of code at the moment. I think I can fix that by creating a view that errors out and then triggering it and then testing the Accept logic.

Anyhow, this is a WIP. I'm curious to see if the approach is daft or not. I also want to see if it passes Travis tests.

@mythmon @rlr Does this seem daft?

willkg:
^^^ That fixes `wsgi/playdoh.wsgi` and `fjord/wsgi.py` so we keep them both and `wsgi/playdoh.wsgi` deals with the server-environment-specific things. This shouldn't hose contributor environments plus it doesn't require us to do anything to stage/prod environments.

I tried to add a test that verifies the Accept handling, but it appears the tests are using a different WSGI mechanism. I might be able to do this with a LiveServerTestCase, but that seems like a lot of trouble.

It's easy to test manually.

1. Run `./manage.py runserver`
2. Go to http://localhost:8000/services/throw-error and see glorious HTML
3. Run this shell script and get text:

```
#!/bin/bash
HOST='http://localhost:8000'

curl -v -XGET "$HOST/services/throw-error" \
     -H 'Accept: application/json'
```

willkg:
^^^ That cleans some things up and reduces some of the differences.

willkg:
I spent some time trying to get LiveServerTestCase working, but it refuses to be ok with `DEBUG=True`, so I can't test the error handling automatedly.

willkg:
For the record, I don't like parts of this PR, but can't figure out better ways to do them and I need this (or something equivalent that lets me see POST payloads) asap so I can fix HB issues.

wsgi/playdoh.wsgi
```diff
@@ -24,18 +28,23 @@ wsgidir = os.path.dirname(__file__)
...
+# application = get_wsgi_application()
+
+from fjord.wsgi import get_debuggable_wsgi_application
+application = get_debuggable_wsgi_application()
+
 
 if newrelic:
     application = newrelic.agent.wsgi_application()(application)
-
-# vim: ft=python
```
rlr: why you hating on vim? &lt;/troll&gt;

fjord/wsgi_utils.py
```diff
@@ -0,0 +1,87 @@
...
+        # Second, check the Accept header and if it's not text/html,
+        # pretend this is an AJAX request so that we get the output in
+        # text rather than html when DEBUG=True.
+        if settings.DEBUG:
+            # request.is_ajax() == True will push this into doing text
+            # instead of html which is waaaaaayyy more useful from an
+            # API perspective. So if the Accept header is anything other
+            # than html, we'll say it's an ajax request to return text.
+            if 'html' not in request.META.get('HTTP_ACCEPT', 'text/html'):
+                request.META['HTTP_X_REQUESTED_WITH'] = 'XMLHttpRequest'
```
rlr: crazypants. I can't believe the rest framework doesnt do something nicer with exceptions.

wsgi/playdoh.wsgi
```diff
@@ -15,7 +20,6 @@ if newrelic:
         newrelic = False
 
 
-os.environ.setdefault('CELERY_LOADER', 'django')
```
rlr: you removed this but I see it in the other wsgi file.

rlr:
Slightly related, it would be cool if that error page was CSRF exempt... So I could do this `$ curl -v -XPOST "http://0.0.0.0:8000/services/throw-error-H 'Accept: application/json' -d "this is the body"`

wsgi/playdoh.wsgi
willkg: Yeah. I moved this to the "common wsgi setup".

fjord/wsgi_utils.py
willkg: This is the "top level uncaught exception" handler which DRF doesn't have anything to do with. I was kind of hoping that DRF would have some helper utilities or something to make this better, but it doesn't.

To be honest, I'm puzzled by this whole affair. How do people debug unhandled API problems if they can't see the POST data at all?

willkg:
I can make that view csrf-exempt. That's a good idea.

wsgi/playdoh.wsgi
willkg: I've been removing these emacs/vim metadata things where I see them.

rlr:
cool. Time to try it on stage?

willkg:
^^^ Added the `csrf_exempt` thing.

rlr:
I see `HTTP_X_POST_BODY = 'this is the body'` now. YAY!

willkg:
Ok. After Travis gives me the green light, I'll squish everything down, double-check the code, push it to upstream in a branch and test it out on stage.

willkg:
^ Squashed commits and redid the commit message.

I'm pushing this to stage to test it out now.

willkg:
I pushed it to stage and tested it out. I tried a few "bad post data" attempts, too, and they all seem fine.

willkg:
Landed in ceb53eb [bug 1136840] Fix error handling for better debugging

Thank you!

PR #510, opened 2015-03-05
[bug 1139545] Handle all exceptions

willkg:
This adds some additional defensive programming to handle all possible
exceptions for the postbody thing.

r?


willkg:
I have no tests for this, but theoretically it should be ok. The HTTP_X_POSTBODY field would have a value that includes the exception repr so we can (over time) handle those issues, too.

rlr:
ooh r+

willkg:
92ce3b6 [bug 1139545] Handle all exceptions

Thank you!

PR #515, opened 2015-03-05
[bug 1139713] Add received_ts field to hb Answer model

willkg:
* add a received_ts field to the hb Answer model with a silly
  initial default value
* backfill the field with the data from updated_ts

r?

### DIFF
--- a/fjord/alerts/models.py
+++ b/fjord/alerts/models.py
@@ -1,12 +1,9 @@
-from django.conf import settings
-from django.core.exceptions import ValidationError
 from django.db import models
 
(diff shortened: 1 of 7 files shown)


willkg:
This re-creates pull #512 which I closed by inadvertently deleting the underlying branch. :p

rlr:
I was a little confused about updated vs received but looked at the bug comments and got it.

r+!

willkg:
ff9f01c [bug 1139713] Add received_ts field to hb Answer model

Thank you!

PR #517, opened 2015-03-05
[bug 1048462] Upgrade django-multidb-router to 0.6

willkg:
r?

Also, sorry I'm doing these one at a time, but I'm pushing them out one at a time so as to reduce "omg! what happened" factor.


rlr:
verified you got the right sha for that release in vendor. r+

willkg:
ae5cb13 [bug 1048462] Upgrade django-multidb-router to 0.6

Thank you!

PR #518, opened 2015-03-06
[bug 1048462] Nix django-compressor

willkg:
We weren't using django-compressor, except for a static finder which
turns out to be mostly a no-op. So we snuck out in the still of the
night and stole the secret plans for the one that kitsune has. NO ONE
WILL DENY US OUR STATIC FILES!

r?

### DIFF
--- a/fjord/base/static_utils.py
+++ b/fjord/base/static_utils.py
@@ -1,0 +1,16 @@
+from django.contrib.staticfiles.finders import BaseStorageFinder
+from django.contrib.staticfiles.storage import StaticFilesStorage
+
+
+class WTFinder(BaseStorageFinder):
+    """A staticfiles finder that looks in STATIC_ROOT.
+
+    This is super lame!
+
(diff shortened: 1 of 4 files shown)


fjord/base/static_utils.py
```diff
@@ -0,0 +1,16 @@
+from django.contrib.staticfiles.finders import BaseStorageFinder
+from django.contrib.staticfiles.storage import StaticFilesStorage
+
+
+class WTFinder(BaseStorageFinder):
```
rlr: HA!!!

rlr:
I nuked the top stati/c dir and loaded the site via devserver and stuff works. WFM
r+

fjord/base/static_utils.py
willkg: I stole that from kitsune verbatim. I think you wrote it!

willkg:
649b136 [bug 1048462] Nix django-compressor

Thank you!

PR #522, opened 2015-03-09
[bug 1139545] Wrap hb integrity error with logging

willkg:
Pretty sure we can fix this just by wrapping it with a transaction, but
I still can't see what's going on. So for now, I'm going to log errors
so I can see what's going on.

r?

### DIFF
--- a/fjord/heartbeat/api_views.py
+++ b/fjord/heartbeat/api_views.py
@@ -1,3 +1,5 @@
+from django.db import IntegrityError
+
 import rest_framework.views
 import rest_framework.response
 from statsd import statsd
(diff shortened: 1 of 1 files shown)


willkg:
I can't reproduce the issue without serious goofypants stuff that's so goofy that I'm not sure it represents the actual issue.

This code logs the post data which is the part I can't see with the error emails so I can see what's going on and can fix the thing we want to fix rather than today's latest unsubstantiated theory.

Anyhow, eyeballing the changes is good. None of the tests go through this.

rlr:
lgtm. go for it!

willkg:
12940b0 [bug 1139545] Wrap hb integrity error with logging

Thank you!

PR #523, opened 2015-03-10
[bug 1139545] Fix "get or create" section of HB post view

willkg:
This tweaks the POST view so that it tries to create the object first.
That will fail with an IntegrityError because there's a uniqueness
constraint on (person_id, survey_id, flow_id). If it fails, then
there's already a record like this and it fetches that and updates it.

In this way, we use the db to eliminate the race condition between
the "get" and "create" steps which is kicking up the IntegrityErrors
we're seeing.

r?

### DIFF
--- a/fjord/heartbeat/api_views.py
+++ b/fjord/heartbeat/api_views.py
@@ -1,4 +1,4 @@
-from django.db import IntegrityError
+from django.db import IntegrityError, transaction
 
 import rest_framework.views
 import rest_framework.response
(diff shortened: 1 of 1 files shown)


willkg:
There's a lot of stuff going on here including the fact that I think I'm right about all this, but I'm not 100% positive.

First off, the problem we're seeing is that an HB client will fire off an HTTP POST and 17 milliseconds later (yes--holy shit that's fast) fires off a second HTTP POST (this is expected behavior except for the "I'm going to send two things 17 ms apart" which is surprising). Whether it's threads or webnodes or the Pirates of Penzance, the two HTTP POSTS get handled by separate things at the "same time".

In the original code, we do a "get or create" kind of thing. We're using Django defaults for transaction management (pretty sure this is true) so it's got autocommit on. So when handler R2D2 does a get, it gets a DoesNotExist, so it moves on to create the object some (short) amount of time later. Meanwhile, handler C3P0 does a get, gets a DoesNotExist and then also moves on to create the object some (short) amount of time later. One of them wins and creates the object and the other throws an IntegrityError. JarJar says, "meesa send admins an eeeemail!" and sends a totally useless email that doesn't have any POST data information.

In the new code, we invert the check and do a "create or get" kind of thing. Using the same characters as above, this means R2D2 and C3P0 do creates at the same time. One wins and the other throws an IntegrityError forcing the loser to go through the update code. JarJar sits at home eating bugs. Will doesn't get useless email. Generally, the universe is a better place.

We have to do the `transaction.atomic()` context because the transaction mode we're in doesn't let us create something and then do queries for it. I was getting this error:

```
TransactionManagementError: An error occurred in the current transaction.
You can't execute queries until the end of the 'atomic' block, jerk!
```

The tests pass, so we're doing the transaction stuff right. I'm pretty sure this will fix the problem we're having in production.

Can someone double-check all the logic and code?

Relevant links:

* https://docs.djangoproject.com/en/1.7/topics/db/transactions/#autocommit-details
* https://docs.djangoproject.com/en/1.7/topics/db/transactions/#django.db.transaction.atomic

NB: My other little one puked while I was in the middle of this PR comment. I don't know if he saw my code or not.

rlr:
yay! Wow, this took me back to like 7 years ago I had a similar race condition in .NET land. GAH! thanks for that. :+1: r+

willkg:
@rlr and I talked on IRC about the atomic section part and I decided to revise it a bit.

^^^ That tweaks that try/except block in respects to the atomic context.

rlr:
That looks like what the docs suggest. r+!

fingers crossed that this brings world peace

willkg:
8b61f14 [bug 1139545] Fix "get or create" section of HB post view

Thank you! (fingers crossed)

PR #524, opened 2015-03-11
[bug 1141410]Fix unicode exception when feedback has invalid unicode URL

lgp171188:


fjord/feedback/tests/test_api.py
```diff
@@ -1,3 +1,5 @@
+# This Python file uses the following encoding: utf-8
+
```
lgp171188: Had to set the encoding of the file to avoid an exception due to the presence of Unicode characters in this file

fjord/feedback/tests/test_api.py
```diff
@@ -366,6 +368,34 @@ def test_maximal(self):
...
+            'user_agent': (
+                'Mozilla/5.0 (Mobile; rv:18.0) Gecko/18.0 Firefox/18.0'
+            ),
+            'source': 'email',
+            'campaign': 'email_test',
+        }
+        try:
+            r = self.client.post(reverse('feedback-api'), data)
+            eq_(r.status_code, 400)
+        except UnicodeEncodeError:
```
lgp171188: If the issue is not fixed, this POST request will cause an exception which will cause the test to fail. So I am not sure if I have to enclose the test in a try-except block and assert an error if there is an exception.

fjord/feedback/tests/test_api.py
willkg: This isn't how you set the encoding. You need to do this:

```
# -*- coding: utf-8 -*-
```

You can see examples in the migration files.

fjord/feedback/tests/test_api.py
willkg: You don't need to catch the error and assert False. Just let the exception propagate.

fjord/feedback/tests/test_api.py
lgp171188: I used one of the ways documented in pep 263 - "Without interpreter line, using plain text" since the style you have quoted was mentioned under "With interpreter binary and using Emacs style file encoding" and I was not sure whether to use it without the interpreter binary shebang line. I will fix it now.

fjord/feedback/tests/test_api.py
lgp171188: Okay will do that.

fjord/feedback/tests/test_api.py
willkg: What I suggested is the standard way to do it since it's recognized by many editors including but not limited to Emacs.

I'm not sure I've seen the other styles. I didn't even know there were other ways to specify the file encoding.

lgp171188:
Fixed the issues

fjord/feedback/tests/test_api.py
```diff
@@ -366,6 +367,31 @@ def test_maximal(self):
...
+            'device': 'OmniCorp',
+            'country': 'US',
+            'user_agent': (
+                'Mozilla/5.0 (Mobile; rv:18.0) Gecko/18.0 Firefox/18.0'
+            ),
+            'source': 'email',
+            'campaign': 'email_test',
+        }
+        r = self.client.post(reverse('feedback-api'), data)
+        eq_(r.status_code, 400)
```
willkg: We should also make sure that we get the "xyz is not a valid url" error message. That way we know the error that was kicked up is the one we want to be kicked up.

willkg:
One minor last thing and then this is good to go.

lgp171188:
Add the test for the response content. Couldn't directly test equality since I couldn't get the expected string to match with the returned string probably because it has Unicode characters and I was doing something wrong. That said, the tests I have written still check that we get an "invalid url" error

willkg:
7fa256a [bug 1141410] Fix unicode exception when feedback has invalid unicode URL

Thank you!

PR #526, opened 2015-03-18
[bug 1144476] stop logging update_ts errors

lgp171188:
Add a log_errors flag (defaults to True) to the HeartbeatV2API.rest_error()
method which disables error logging when set to False. Use it to not log
the error when the update_ts error happens.


willkg:
Looks good to me. Thank you!

willkg:
026d9ae [bug 1144476] stop logging update_ts errors

PR #529, opened 2015-03-19
[bug 1038512] Rewrite translation backfill admin page

willkg:
The previous backfill page pulled all the instances for a given model
and date range (ouch) and then for each instance, told it to go create
translation tasks (double-ouch). That's a ton of work and far exceeds
the amount of time alotted for connections.

This version reimplements that and pushed the work to celery tasks.

r?

### DIFF
--- a/fjord/feedback/models.py
+++ b/fjord/feedback/models.py
@@ -121,6 +121,22 @@ class Product(ModelBase):
         return self.__unicode__().encode('ascii', 'ignore')
 
 
+class ResponseManager(models.Manager):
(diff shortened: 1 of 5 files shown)


willkg:
The important bits are covered with a couple of tests. The admin page itself is not, but that's fine because it's an admin page we don't use very often.

rlr:
I eyeballed this and it looks good. Tests pass. r+

willkg:
5c61ffc [bug 1038512] Rewrite translation backfill admin page

Thank you!

PR #531, opened 2015-03-20
[bug 1144870] Add trans command

willkg:
This gives me a command-line way to test translation systems. It's a lot
easier than creating feedback and then triggering a translation for it
and then removing the translation from the db.

This just adds a command and tweaks some code to work better. It doesn't affect tests or production or anything else.

r?

### DIFF
--- a/fjord/journal/models.py
+++ b/fjord/journal/models.py
@@ -25,7 +25,7 @@ class RecordManager(models.Manager):
             msg=msg,
             metadata=metadata
         )
-        if instance:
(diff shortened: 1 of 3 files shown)


willkg:
First question would be "why call it 'trans' instead of 'translate'"? and that's because dennis django stuff already uses translate. I think I might stop using that, though, but have to wait for the requirements/vendor situation to be better.

rlr:
LGTM r+

willkg:
9e7f26d [bug 1144870] Add trans command

Thank you!

PR #532, opened 2015-03-20
[bug 1137022] Overhaul settings and cleanup startup

willkg:
* nix fjord/base/monkeypatches.py and fold the code into
  fjord/manage_utils.py
* clean up startup
* rework wsgi/playdoh.wsgi so it no longer imports manage.py for
  side-effects (gross) but instead explicitly sets up the environment
  properly
* add settings_utils.py and infrastructure we can later use for settings that come from the environment

r?


willkg:
^^^ That rewrites everything so far.

Instead of making all the changes to settings files, this reworks Fjord startup to clean it up a bunch, cleans up some related code and then adds `settings_utils.py` which has the functions we can use for settings that need to pull from the environment.

For now, I think we should just leave it like it is. We'll continue to use the existing settings system where you have `base.py` and `local.py` overrides things so we don't incur the pain of having two different settings systems co-existing for different environments like we're currently doing with requirements/vendor. Moving to AWS is months away, so I think it's better to not add complexity now.

willkg:
This is ready for review.

After it passes Travis and review, I'll test it out on stage. I did make changes to playdoh.wsgi, but I suspect it'll either work or not work and be pretty obvious. We can fix those issues then. Then we'll land it in master and push to stage and prod.

manage.py
```diff
@@ -1,24 +1,23 @@
...
 
+    # This has to get called after DJANGO_SETTINGS_MODULE stuff has
+    # been sorted out.
+    manage_utils.setup_environ()
+    manage_utils.monkeypatch()
 
-if USING_VENDOR == '0' and SKIP_CHECK != '1':
-    manage_utils.check_dependencies()
+    SKIP_CHECK = os.environ.get('SKIP_CHECK', '0')
+    USING_VENDOR = os.environ.get('USING_VENDOR', '0')
```
mythmon: It seems like the default for this used to be `'1'`. Now it is `'0'`. Are you making changes elsewhere in prod to make this ok?

willkg: The diff for this file *sucks*. This code just got moved and indented. It's otherwise identical. The defaults are the same.

mythmon: Sorry, what I was actually pointing out was that this sets a default, and in playdoh.wsgi you used to change that. That doesn't happen anymore in playdoh.wsgi.

wsgi/playdoh.wsgi
```diff
@@ -20,25 +20,22 @@ if newrelic:
...
+# environment already.
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'fjord.settings')
 
 # Add the app dir to the python path so we can import manage.
 wsgidir = os.path.dirname(__file__)
 site.addsitedir(os.path.abspath(os.path.join(wsgidir, '../')))
 
-# Explicitly set these so that fjord.manage_utils does the right
-# thing in production.
-os.environ['USING_VENDOR'] = '1'
```
mythmon: Does this mean production needs to change somehow?

willkg: It doesn't need to change in production because this no longer imports `manage.py` for side-effects. Instead it does the things it needs to do explicitly. i.e. `setup_environ()` and `monkeypatch()`.

mythmon: Oh, I see.

manage.py
```diff
@@ -1,24 +1,23 @@
...
-
-from fjord.base import monkeypatches
-monkeypatches.patch()
 
+if __name__ == "__main__":
+    os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'fjord.settings')
 
-SKIP_CHECK = os.environ.get('SKIP_CHECK', '0')
-USING_VENDOR = os.environ.get('USING_VENDOR', '0')
+    print 'Using %s' % os.environ['DJANGO_SETTINGS_MODULE']
```
mythmon: This printed out twice for me. Is that intended/known?

mythmon:
It seems weird to me to add a bunch of dead code like this, even though it makes sense. I would personally start putting `settings_utils.config` calls in `fjord/settings/base.py`. The concerns you raised on IRC about those getting tested or not make sense though.

The changes to start up make sense, and the helpers you wrote for env-based settings seem sound.

manage.py
willkg: It prints for me twice, too. I think what's going on is that manage.py gets run by the test setup. I'll investigate that now.

willkg: Scratch that--I only see it once now. I'm not sure why you'd be seeing it twice. Maybe tweak it so that it prints the sys.argv, too, so then you can see whether manage.py is being executed twice with different args?

willkg:
^^^ That adds some uses of `settings_utils.py` things. Pretty sure they're all overridden in travis setup or `local.py`, so I think it's moot, but they're there.

manage.py
mythmon: I figured out that `./manage.py runserver` spawns another process, one for web serving another file file watching. It is goofy, but not a lot that we can do.

mythmon:
This looks good to me, though if you have any clever ideas on how to fix that double "using fjord.settings" messages, that would be cool.

I'd really like to see what Travis says about this, but r+ from me.

willkg:
^^^ That tweaks the list issues in config.

So, when you run the server with `./manage.py runserver`, it actually tells you the settings being used:

```
Performing system checks...

System check identified no issues (0 silenced).
March 24, 2015 - 14:34:23
Django version 1.7.7, using settings 'fjord.settings'                  <<<
Starting development server at http://127.0.0.1:8000/
Quit the server with CONTROL-C.
```

So I'll just nix that print line.

willkg:
^^^ Ok. I think that covers it. I'll test on stage and report back if there are issues.

Thank you!

willkg:
Looks good on stage. I think this is good to go.

willkg:
4bb85f8 Nix unneeded print line
e4e615d [bug 1137022] Add settings_utils.py uses in base.py
84b11d8 [bug 1137022] Rework startup; add settings_utils.py

Thank you!

mythmon:
Oh. The error that Travis got is the same thing I got on my laptop. I figured it was unrelated, but I guess not?

willkg:
I think it's from one of the things we config'd. I'll look into it.

willkg:
Should be fixed in pull #536.

PR #536, opened 2015-03-25
[bug 1137022] Fix CACHES setting in tests

willkg:
This should fix the travis issue we had.

### DIFF
--- a/fjord/settings/base.py
+++ b/fjord/settings/base.py
@@ -4,7 +4,6 @@
 
 import logging
 import os
-import socket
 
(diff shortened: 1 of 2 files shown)


willkg:
^^^ @mythmon That should fix your issue, too. I had CACHES set in my local.py file, so it worked for me previously.

mythmon:
This looks good, and fixes the problem with the tests for both me and Travis.

r+

willkg:
9a95852 [bug 1137022] Fix CACHES setting in tests

Thank you!

PR #537, opened 2015-03-25
[bug 1147227] Add notes about frames

willkg:
@mythmon Can you eyeball these comments? Is the language correct?

### DIFF
--- a/fjord/feedback/tests/test_views.py
+++ b/fjord/feedback/tests/test_views.py
@@ -26,6 +26,18 @@ class TestRedirectFeedback(TestCase):
 class TestFeedback(TestCase):
     client_class = LocalizingClient
 
+    def test_xframe_header(self):
+        """Feedback form should *always* have X-FRAME-OPTIONS: DENY
+
+        This is because the feedback form can do various magic things
+        that shouldn't ever be framed.
+
+        """
+        url = reverse('feedback', args=(u'firefox',))
+        resp = self.client.get(url)
+        eq_(resp.status_code, 200)
+        eq_(resp['X-Frame-Options'], 'DENY')
+
     def test_valid_happy(self):
         """Submitting a valid happy form creates an item in the DB.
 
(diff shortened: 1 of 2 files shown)


mythmon:
Sounds good to me.

I think the Travis failure is related to the cache problem you fixed recently?

willkg:
Yes. The travis failure is fixed in pull #536.

mythmon:
Cool. r+

willkg:
I'm going to change this around and instead of adding the comment to settings, I'll add a test to make sure the header is there for the feedback form.

willkg:
^^^ That changes the notes a bit and also adds a test.

r?

fjord/feedback/tests/test_views.py
```diff
@@ -26,6 +26,17 @@ def test_sad_redirect(self):
...
+    def test_xframe_header(self):
+        """Feedback form should *always* have X-FRAME-OPTIONS: DENY
+
+        This is because the feedback form can do various magic things
+        that shouldn't ever be framed.
+
+        """
+        url = reverse('feedback', args=(u'firefox',))
+        resp = self.client.get(url)
+        eq_(resp['X-Frame-Options'], 'DENY')
```
mythmon: I'd also check for 200 OK, since your 40x/50x page probably has different X-Frame-Options properties.

mythmon:
I'd be more comfortable if this was rebased onto master to pick up that cache config fix so Travis could run this test. But the test looks right. I'd suggest making it a little more defensive too though.

willkg:
I can do that, but that hasn't been reviewed, yet.

willkg:
^^^ Rebases against master with the CACHES fix. I'll keep an eye on Travis.

mythmon:
Travis passed it. r+

willkg:
7e796de [bug 1147227] Add note to feedback view and test re: X-FRAME-OPTIONS

Thank you!

PR #539, opened 2015-03-26
[bug 1127428] Uncomment booleanfield test

willkg:
The BooleanField required=True problem is gone now. I uncommented out
the test, but then bumped into a goofy edge case where BooleanField
defaults to False if you do an HTML form post.

So I went through all the test cases and converted them all to JSON
payloads. That fixes this test and is more correct test-wise since our
API docs ask people to use JSON payloads.

Quick r?

### DIFF
--- a/fjord/feedback/tests/test_api.py
+++ b/fjord/feedback/tests/test_api.py
@@ -358,7 +358,10 @@ class PostFeedbackAPITest(TestCase):
             'product': u'Firefox OS'
         }
 
-        r = self.client.post(reverse('feedback-api'), data)
+        r = self.client.post(
+            reverse('feedback-api'),
+            content_type='application/json',
+            data=json.dumps(data))
         eq_(r.status_code, 201)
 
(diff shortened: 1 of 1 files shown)


fjord/feedback/tests/test_api.py
```diff
@@ -580,26 +605,28 @@ def test_invalid_email_address_returns_400(self):
...
+            'product': u'Firefox OS',
+            'locale': 'en-US',
+        }
+
+        r = self.client.post(
+            reverse('feedback-api'),
+            content_type='application/json',
+            data=json.dumps(data))
+        eq_(r.status_code, 400)
+        assert 'happy' in r.content
```
willkg: This is the test we uncommented. Works super now.

mythmon:
I read the uncommented test and skimmed over the rest. Looks good.

r+

willkg:
Landed in 47db8d6 [bug 1127428] Uncomment booleanfield test

Thank you!

PR #540, opened 2015-03-30
[bug 1141056] Fix history in onpopstate

willkg:
When onpopstate kicks off, we were changing the page, but also tweaking
the history. We don't need to do that last bit since the event only
fires on history changing.

Quick r?


rlr:
gah! r+ when travis goes :green_heart:

willkg:
6352e56 [bug 1141056] Fix history in onpopstate

Thank you!

PR #541, opened 2015-03-30
1143881 prefill

willkg:
r?

### DIFF
--- a/fjord/base/static/js/fjord_utils.js
+++ b/fjord/base/static/js/fjord_utils.js
@@ -52,8 +52,6 @@ window.fjord = window.fjord || {};
             var opts = qs.split('&');
             for (i=0; i < opts.length; i++) {
                 var opt = opts[i].split('=');
-                console.log(opts[i]);
-                console.log(opt);
                 parsed[opt[0]] = opt[1];
             }
         }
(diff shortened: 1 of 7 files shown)


willkg:
To test by hand:

1. go to http://localhost:8000/feedback/?happy=0
2. pick a product, make sure you land on the more-info page
3. submit
4. make sure the resulting response is sad

Also this to make sure the user can override the prefill:

1. go to http://localhost:8000/feedback/?happy=0
2. pick a product and make sure you land on the more-info page
3. click on the browser back button and make sure you go back to the sentiment page
4. pick "happy"
5. fill out a description and submit
6. make sure the resulting response is happy

You can also run the smoketests, but if they're not working well for you then it's 3 minutes of noise.

rlr:
Do you need to do something additional for FxOS since it's a separate form? It doesn't seem to handle it for me.

rlr:
Other than that, this is working. If that is expected, r+

willkg:
My bad--I should have mentioned that I didn't implement this for Firefox OS--just for all the non-Firefox OS products. I'm not sure this is necessary for Firefox OS, so I just left it out for now.

willkg:
5eaa494 [bug 1143881] Add happy pre-filling via url

Thank you!

PR #542, opened 2015-03-30
Fix DATABASES and CACHES settings in local.py-dist

willkg:
This makes it easier to set up development locally.

@rlr Does this look ok? r?


rlr:
I likey. r+!

willkg:
b33bf9b Fix DATABASES and CACHES settings in local.py-dist

PR #544, opened 2015-03-30
[bug 910789] Redo sampledata and add api responses

willkg:
This redoes how we generate sample data for feedback responses:

1. adds api responses based on the product
2. changes it to 100 happy + 100 sad + 1000 other responses
3. fixes it so we're looking at product tuples which group a product,
   version, platform and useragent together -- previously we did
   them separately which resulted in weirdo/impossible combinations
4. adjust the locale generator so it looks a bit more like typical
   Input data

This might also fix problems with the smoketests.

r?

### DIFF
--- a/fjord/feedback/sampledata.py
+++ b/fjord/feedback/sampledata.py
@@ -48,55 +48,48 @@ URLS = (
     u'http://mozilla.org/',
 )
 
-PRODUCTS = (
-    (u'Firefox', u'29.0'),
-    (u'Firefox', u'28.0'),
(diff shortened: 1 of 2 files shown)


rlr:
WFM. I still get the same smoke test error I usually get (test_date_filter). r+

willkg:
I'll have to look into that. Maybe generating 1000 isn't enough. Maybe they need to be closer together datetime-wise.

7e6defd [bug 910789] Redo sampledata and add api responses

Thank you!

PR #545, opened 2015-04-15
[bug 1149517] Restyle thank you page

willkg:
This restyles the thank you page to match the generic feedback form
flow. It's now responsive and works on smaller screens. It doesn't
change any of the strings.

r?

### DIFF
--- a/fjord/base/static/css/fjord.less
+++ b/fjord/base/static/css/fjord.less
@@ -306,12 +306,12 @@ footer {
   .contents {
     text-align: left;
     margin: auto;
-    max-width: 940px;
+    max-width: 700px;
 
     clear: both;
     background: url(../img/footer-logo.png) top left no-repeat;
     line-height: 140%;
-    padding-left: 60px;
+    padding: 0 30px 0 60px;
 
     p {
       margin-bottom: 0.5em;
@@ -340,8 +340,15 @@ footer {
     }
   }
 
-  #footer-links {
+  ul#footer-links {
     font-size: 85%;
+    padding: 0;
+    li {
+      display: inline;
+      &:not(:last-child):after {
+        content: " | ";
+      }
+    }
   }
 }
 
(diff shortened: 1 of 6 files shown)


willkg:
Things to know:

1. This redoes the style of the Thank You page so that it's the same style as the rest of the generic feedback form flow. Gone is the shock of filling out the feedback form and then feeling like you ended up on a different site.
2. The rest of the feedback form follows a "mobile-first" approach where the default CSS is geared towards small viewports and then there are media queries for viewports with larger widths. This brings the Thank You page inline with that. Further, now the Thank You page is functional on mobile devices--previously it sucked.
3. These changes aren't covered by the unit tests.

To test:

1. `./manage.py runserver`
2. open up the site in your browser and go to `/feedback/`
3. fill out the feedback form and get to the Thank You page
4. switch to responsive mode in your browser and view at 320px (smallest viewport we handle), 768px and bigger than that --- all three should look ok

I'm definitely interested in thoughts on how to make it better. I'm not wild about what I did with the "follow us" icons in 320px. Things feel out of alignment in respects to the footer links. But it kind of feels minor.

Further, if you're looking at it with a width 768 or greater, there's a lot of empty space on the right side. I thought about bringing in the margins, but that didn't feel any better.

I think both of these are minor. I think the form is fine as is. There are probably a lot of other icky feels in the style of the site, so I don't think this sticks out like a sore thumb.

rlr:
1000 times better. What bothers me the most is the sharing icons not vertically aligned with the text. I played with it but not sure how to center align them.

rlr:
That said, this is way better. It can be fine polished from here but not sure if there is much value in doing it. r+

willkg:
I'd love to have the icons center aligned. I couldn't figure it out either.

willkg:
NB: The Travis failures are from ssl connection errors when peep was trying to download one of the python requirements. Thus some of the Python requirements didn't get installed. Thus fail.

PR #546, opened 2015-04-20
[bug 1087571] Codify editor settings using editorconfig

lgp171188:
Add the settings currently used by kitsune. Can add more specific settings later if required.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -12,6 +12,10 @@ charset = utf-8
 indent_style = space
 indent_size = 4
 
+[*.less]
+indent_style = space
+indent_size = 2
+
 [{package.json,.travis.yml}]
 indent_style = space
 indent_size = 2
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# http://editorconfig.org
+
+root = true
+
+[*]
+insert_final_newline = true
+trim_trailing_whitespace = true
+end_of_line = lf
+charset = utf-8
+
+[*.{js,py}]
+indent_style = space
+indent_size = 4
+
+[{package.json,.travis.yml}]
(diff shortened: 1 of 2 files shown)


.editorconfig
```diff
@@ -0,0 +1,15 @@
+# http://editorconfig.org
+
+root = true
+
+[*]
+indent_style = space
+indent_size = 4
```
willkg: So, this is where it gets tricky. We use 4-space indents for .py and .js files. For .less files we use 2-space indents. I don't know offhand whether there are other rules, but I think we should move the indent bits to a `[*.py,*.js]` block and add this:

```
[*.less]
indent_style = space
indent_size = 2
```

willkg:
Other than that thing about `indent_size`, this looks great.

We probably want to mention something in the docs so that contributors can set up editorconfig things in their editor of choice. It's probably sufficient to put it in the Getting Started guide and link to the part of the editorconfig site where they mention editor plugins.

This looks great! Thank you!

lgp171188:
Implemented the changes suggested. Let me know if it is okay @willkg

.editorconfig
```diff
@@ -0,0 +1,17 @@
...
+[*]
+insert_final_newline = true
+trim_trailing_whitespace = true
+end_of_line = lf
+charset = utf-8
+
+[*.{js,py}]
+indent_style = space
+indent_size = 4
+
```
willkg: This needs one more section:

```
[*.less]
indent_style = space
indent_size = 2
```

lgp171188: Forgot to add it in my previous commit. Added it now

willkg:
Looks great! Thank you!

PR #548, opened 2015-04-20
[bug 1156388] Fix EDITOR req in commit msg linter

willkg:
r?

### DIFF
--- a/bin/lint_commit_msg.py
+++ b/bin/lint_commit_msg.py
@@ -4,6 +4,7 @@ import os
 import re
 import subprocess
 import sys
+import textwrap
 
 import requests
 from requests.exceptions import (
@@ -12,6 +13,14 @@ from requests.exceptions import (
 )
 
 BUG_PREFIX_REGEX = r'\[bug (\d+)\]'
+
+
+def wrap_paragraphs(text):
+    paragraphs = [
+        '\n'.join(textwrap.wrap(paragraph))
+        for paragraph in text.split('\n')
+    ]
+    return '\n'.join(paragraphs)
 
 
 def are_lines_not_more_than_79_chars(contents):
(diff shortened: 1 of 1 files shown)


willkg:
@lgp171188 Does this look ok to you?

lgp171188:
I hadn't understood what you had meant when you told on #input -  "We will treat it as if the user had entered 'i' ". I would say that we should abort the commit if EDITOR is not set and let the user redo it instead of ignoring the issue without the user explicitly wanting it.

bin/lint_commit_msg.py
```diff
@@ -104,6 +105,19 @@ def lint_commit_msg(commit_msg_file, editor):
         print ''
         print '\n'.join(errors)
         print ''
+        if not editor:
+            print (
+                '\n'.join(textwrap.wrap(
+                    'You do not have EDITOR set in your environment. Please '
+                    'set EDITOR in your environment if you want to be able '
+                    'to edit the commit message. In the meantime, you can do:'
```
lgp171188: I would say this sounds better - "Please set EDITOR in your environment if you want to be able to edit the commit message during the commit if it does not conform to the recommended format. Otherwise you can do ```git commit --amend``` to edit the commit message.

bin/lint_commit_msg.py
```diff
@@ -104,6 +105,19 @@ def lint_commit_msg(commit_msg_file, editor):
...
+                )) +
+                '\n\n'
+                'git commit --amend\n\n'
+                'to fix issues.'
+            )
+            return 0
```
lgp171188: I hadn't understood what you had meant when you told on #input - "We will treat it as if the user had entered 'i' ". Now after looking at the code I do :-)

I would say that we should not return 0 and let the commit succeed if EDITOR is not set. This is because we check the presence of EDITOR environment variable only if there are issues with the commit message. 

Instead we could prompt the user with the y/n/i choice and let him choose what to do. We can also enhance the description of the "i" option to say "the commit will succeed as is, I will edit the commit message manually if needed" and that of "n" option to say "the commit will fail, I will redo it"

willkg: One of the things I fixed in the previous pass that kept biting me was the problem where the commit message effectively disappears if the linter rejects it forcing me to retype the whole thing from scratch. Technically it's in the file `.git/COMMIT-MSG`, but once you start a new commit, that file disappears. So you have to copy the contents of that file before trying the commit again. It's pretty clunky and I don't like re-typing long commit messages.

I agree with what you're thinking here, but I want to make it easy to fix the commit message if there's no `EDITOR` set without having to retype it from scratch. If we simulate an "ignore" here and let the commit message through, then the instructions for fixing the commit message are much easier to walk through.

I'm game for changing the logic here and having a choice with a "no" that causes the linter to reject the message if there's an easy way to allow the user to edit the message rather than having to do it from scratch. Do you know if that's possible?

willkg:
^^^ That tweaks the output a bit to make it clearer. It's not exactly what you suggested, but I think it addresses the point you were trying to clear up.

Does that look ok?

willkg:
^^^ That's the updates commit with your suggested fix to the wording.

Does that look ok?

lgp171188:
Looks perfect. I think this good to be merged

willkg:
Awesome! Thank you!

PR #549, opened 2015-04-23
1156334 restyle fxos

willkg:
Yeah, so that happened.

r?


willkg:
This cleans up a bunch of stuff.

There's still a lot of duplicate js code between the generic and fxos forms, but they're pretty different under the hood because they had different requirements, so I'm going to leave it for now.

This passes the unit tests and also the smoke tests.

Reviewing this might be hard because the diffs are a mess.

To review:

1. look at the final code and comment on issues there
2. do generic and fxos feedback looking for flow issues with back-and-forth using browser back button and the form's back button
3. look at the final css for sanity
4. look at cards.js -- does that look ok?

fjord/feedback/static/js/cards.js
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * - all cards are divs with class "card"
+ * - inactive cards have the "inactive" class; when starting out
+ *   all cards should be marked inactive but the first one
+ * - the back button has the "back-button-container" class
+ * - data-focus attribute on the card has the id of the element
+ *   that should receive focus
+ * - header -> h1 gets the value of the data-title attribute on
+ *   the card
+ */
```
willkg: This sucks. It should probably be a class where the above things are configurable, but I was trying to minimize the number of changes I was doing so I opted to document it for now and maybe fix it later.

fjord/settings/base.py
```diff
@@ -650,7 +637,7 @@ def JINJA_CONFIG():
 ## Logging
 LOG_LEVEL = logging.INFO
 HAS_SYSLOG = True
-SYSLOG_TAG = "http_app_playdoh"  # Change this after you fork.
+SYSLOG_TAG = 'http_app_playdoh'  # Change this after you fork.
```
willkg: I think the pre-commit hooks did this.

mythmon: Seems likely: https://github.com/mozilla/fjord/blob/c115b7a14e06d16ebd06958b9656ec5bb5eb1f59/.pre-commit-config.yaml#L12

willkg:
I need to rebase against master to pick up the footer width fix.

willkg:
^^^ That folds in the footer width fix I just landed. Plus it nixes a bunch of cleanup commits. We can do those in another PR. Hopefully that reduces some of the review work.

fjord/feedback/static/js/fxos_feedback.js
```diff
@@ -1,130 +1,173 @@
...
+        }
+        $('#happy-button').click(function(ev) {
+            selectHappySad(1);
+            return false;
+        });
+        $('#happy-button').keypress(function(ev) {
+            if ((ev.keyCode || ev.which) === '13') {
+                $('#happy-button').click();
+            }
+        });
```
rlr: nit: you could chain these two above. like:
```
    $('#happy-button')
        .click(function(ev) {
            selectHappySad(1);
            return false;
        })
        .keypress(function(ev) {
            if ((ev.keyCode || ev.which) === '13') {
                $('#happy-button').click();
            }
        });
```

It's mostly a styling preference but it saves a dom call to get the node.

fjord/feedback/templates/feedback/fxos_feedback.html
```diff
@@ -13,221 +8,264 @@
...
+            </div>
+
+            <div id="sad-button" class="sentiment-button" tabindex=0>
+              <svg class="face sad-face" viewBox="0 0 100 100">
+                <g>
+                  <!-- 100 by 100 -->
+                  <circle cx="50" cy="50" r="43" stroke-width="7" fill="none"/>
+                  <circle cx="33" cy="37" r="7" stroke-width="0" />
+                  <circle cx="67" cy="37" r="7" stroke-width="0" />
+                  <path d="M24 72 C 35 55, 65 55, 76 72" stroke-width="6" fill="none"/>
```
rlr: i might play some of these numbers on the powerball.

rlr:
I like the structure of the code. It all makes sense. I'm going to do a pass at testing now.

rlr:
I went through the flow for Firefox and Firefox OS. Hit back a bunch of time from the page and from the browser back button. Didn't hit any issues. SHIP IT :ship: r+

fjord/feedback/static/js/fxos_feedback.js
willkg: Oh, neato! Will change!

fjord/feedback/templates/feedback/fxos_feedback.html
willkg: Ha! If you win, you should submit some happy feedback.

PR #550, opened 2015-04-24
[bug 1155905] Fix footer width in thanks page

willkg:
On retina displays, the footer didn't extend all the way to the right
side of the "page". This fixes that.

Quick r?

### DIFF
--- a/fjord/feedback/static/css/generic_feedback.less
+++ b/fjord/feedback/static/css/generic_feedback.less
@@ -407,8 +407,9 @@ aside {
     color: #909090;
     font-size: 11px;
     height: 190px;
-    padding: 0 20px;
+    padding: 0;
     position: absolute;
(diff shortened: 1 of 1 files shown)


willkg:
This won't affect tests. Best way to review is to launch the server and check the footer in the home page and the footer in http://127.0.0.1:8000/en-US/thanks and make sure they look ok.

Pretty sure the problem only happened in retina displays. I "faked" the issue by deleting text in the footer such that the last line was half as long.

See the bug for details and a screenshot of the issue.

willkg:
^^^ Fixed a silly thing that I don't want anyone to ever see so I force-pushed it.

rlr:
I have a retina display, I loaded the thanks page and varied the width and things look fine. r+

willkg:
Thank you!

PR #551, opened 2015-04-24
[bug 1131124] Fix vagrant vm elasticsearch issues

lgp171188:
Create a customized elasticsearch.yml and use it when provisioning the
VM. Move everything related to vagrant inside vagrant/ directory. Update
the RAM of the VM to 1 GB, but it is not working :(


willkg:
I tried to test this out and then hit problems with the virtualenvironment in the virtual machine. Namely, it's failing the dependency check:

```
(fjordvagrant)vagrant@vagrant-ubuntu-trusty-64:~/fjord$ ./manage.py test
Using test configuration
There are 8 requirements that cannot be checked.
The following requirements are not satsifed:

UNSATISFIED: argparse==1.3.0

Update your virtual environment by doing:

    ./peep.sh install -r requirements/requirements.txt
    ./peep.sh install -r requirements/compiled.txt
    ./peep.sh install -r requirements/dev.txt

or run with SKIP_CHECK=1 .
```

When I do `pip freeze`, it shows that argparse 1.2.1 is installed. That's weird because we never specify 1.2.1 anywhere.

So I manually checked the virtual environment and the `PKG-INFO` file says it's got 1.3.0 installed.

If you do:

```
(fjordvagrant)vagrant@vagrant-ubuntu-trusty-64:~/fjord/requirements$ python
Python 2.7.6 (default, Mar 22 2014, 22:59:56) 
[GCC 4.8.2] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> import argparse
>>> argparse.__file__
'/usr/lib/python2.7/argparse.pyc'
>>> 
```

then you can see it and it's loading the argparse module from Python 2.7--not the one installed in the virtual env.

Anyhow, long story short is that there's a lot of weirdness going on and thus `./manage.py` balks on the dependency check and won't go further.

This is with a fresh vm, so I'm thoroughly puzzled.

Do the tests pass for you in a fresh vm created with the vagrant stuff in this PR?

lgp171188:
I see the same issues here. But on reading the pypi page for argparse, I find that it says "As of Python >= 2.7 and >= 3.2, the argparse module is maintained within the Python standard library.". So I have removed it in the most recent commit on this pull request. I tested it and no longer get the argparse error.

lgp171188:
I did a history rewrite to order the commit and fix just the elasticsearch issue here. I will make the argparse changes in a separate pull request. I hope that is fine @willkg

willkg:
This looks ok except VirtualBox thinks the image is a 32-bit OS image.

With richard, VirtualBox thinks the image is a 64-bit OS image which it is. I don't know offhand if this is cosmetic or if it suggests VirtualBox is running 32-bit or something like that.

willkg:
I think you need to rebase this on the top of master tip because that has the argparse changes.

lgp171188:
Rebased it on top of the master branch which should include the argparse change as well.

willkg:
You rock!

willkg:
Awesome! Looks great! Let's land this!

PR #552, opened 2015-04-27
Cleanup commits

willkg:
This cleans up files that aren't used anymore. Plus it copies `thanks.html` to `thanks_sad.html` and removes all the dead thanks page code.

Quick r?


willkg:
To test:

1. run `./manage.py collectstatic` -- that'll error out if there are issues with the bundles and missing files

That's pretty much it. The rest isn't seen by users so even if it was a problem, it's moot.

rlr:
r+ :shipit:

PR #555, opened 2015-04-28
[bug 1158473] Remove simplejson requirement

willkg:
We don't need simplejson anymore, so this removes it. This reduces the
number of possible places that JSONDecodeError exists.

Quick r?


rlr:
r+ when travis goes :green_heart:

willkg:
Thank you!

PR #557, opened 2015-04-29
Upgrade to peep 2.4

rlr:
I ran into issues with peep after updating pip to v6.1.1. This fixes that.

r?


willkg:
I'm mildly irritated we're taking out my comments and replacing them with ed's which i don't think are particularly helpful but are overly verbose. But in the interests of sharing these files with kitsune in a way that's easy, I'm game.

PR #558, opened 2015-04-29
[bug 1159882] Change "happy" in API to default to False

willkg:
Bug #1155600 creates the problem where Firefox OS > 1.3 relies on the
old (busted) behavior where the "happy" field wasn't required and
defaulted to False. This changes the BooleanField back to that behavior
so that Input stops rejecting Firefox OS > 1.3 feedback.

One caveat of this is that all feedback will be "sad" regardless of
whether the user was actually happy or not. That makes me sad, too.

r?

### DIFF
--- a/fjord/feedback/models.py
+++ b/fjord/feedback/models.py
@@ -637,7 +637,14 @@ class PostResponseSerializer(serializers.Serializer):
     emails.
 
     """
-    happy = serializers.BooleanField(required=True)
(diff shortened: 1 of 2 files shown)


willkg:
The changes aren't particularly interesting. The review should be straight-forward.

rlr:
lgtm. r+

willkg:
Thank you!

PR #559, opened 2015-04-30
Move vendor/ path setup to fjord/__init__.py.

rlr:
This is in preparation for celery changes that require code to be in
this file as well and need vendor to be setup.

r?

The tests cover the `manage.py` path. The `wsgi.py` path has to be covered by deploying to -stage?


willkg:
This looks good to me!

PR #561, opened 2015-05-01
Add "./manage.py check" to travis run

willkg:
r?

### DIFF
--- a/bin/travis/dispatch.sh
+++ b/bin/travis/dispatch.sh
@@ -3,11 +3,11 @@
 SUITE=${1:-all}
 
 case $SUITE in
-    all )
-        bin/travis/test.sh
-        ;;
     django )
         bin/travis/test.sh
+        ;;
+    djangocheck )
+        bin/travis/djangocheck.sh
         ;;
     * )
         echo "Unknown test suite '$SUITE'."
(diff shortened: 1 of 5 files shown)


mythmon:
In SUMO, we have a check in the install and run scripts that skip over a lot of the set up when TEST_SUITE=lint to make the tests go faster. Do you want to do something like that here? You'd probably not be able to skip installing Python dependencies. But maybe you could skip installing ES and starting services.

bin/travis/dispatch.sh
```diff
@@ -3,12 +3,11 @@
...
 
 case $SUITE in
-    all )
-        bin/travis/test.sh
-        ;;
     django )
         bin/travis/test.sh
         ;;
+    djangotest )
+        bin/travis/djangocheck.sh
```
mythmon: Oh, you need to have a `;;` after this line. It is like `break` in other languages' case statements. Apparently not having it is a syntax error here (according to Travis)

willkg: Bah. Copy/paste fail.

bin/travis/install.sh
```diff
@@ -12,6 +12,10 @@ echo "Install Python dependencies"
 pip freeze
 echo
 
+if [[ $TEST_SETUP == "minimal" ]]; then
```
mythmon: You'll need a similar check in setup.sh

willkg:
^^^ That fixed all the issues and now passes travis.

mythmon:
LGTM r+

PR #563, opened 2015-05-04
[bug 1161144] Implement suggest framework

willkg:
This implements a suggest framework that allows us to have a series of
Providers which offer suggested links to list on the Thank You page
after a user has given us feedback. Initially, we'll use this to generate
links for sad feedback based on SUMO data. We can additionally use this
for other purposes such as soliciting additional information based on
trigger words.

The world will be our oyster.

r?

### DIFF
--- a/fjord/base/plugin_utils.py
+++ b/fjord/base/plugin_utils.py
@@ -1,0 +1,45 @@
+import logging
+
+from django.utils.module_loading import import_string
+
+
+def load_providers(provider_list, logger=None):
+    """Loads providers in the provided list
+
+    :arg provider_list: list of Python dotted module paths; the provider
+        class should be the last name in the path
+
+    :arg logger: (optional) logger to log to for errors
+
+    :returns: list of providers.
+
+    >>> load_providers(['some.path.Class', 'some.other.path.OtherClass'])
+    [<Class>, <OtherClass>]
+
+    .. Note::
+
+       All exceptions when importing and loading providers are
+       swallowed and logged as errors. Providers that don't import or
+       don't load are not returned in the return list.
+
+    """
+    if logger is None:
+        logger = logging.getLogger('i.plugin_utils')
+
+    provider_objects = []
+    for provider in provider_list:
+        try:
+            # Import the class and instantiate it.
+            provider_objects.append(import_string(provider)())
+        except Exception:
+            logger.exception('Provider {0} failed to import'.format(provider))
+
+    providers = []
+    for provider in provider_objects:
+        try:
+            provider.load()
+            providers.append(provider)
+        except Exception:
+            logger.exception('Provider {0} failed to load'.format(provider))
+
+    return providers
--- a/fjord/suggest/apps.py
+++ b/fjord/suggest/apps.py
@@ -1,0 +1,19 @@
+import logging
+
+from django.apps import AppConfig
+from django.conf import settings
+
+from fjord.suggest import _PROVIDERS
+from fjord.base.plugin_utils import load_providers
+
+
+logger = logging.getLogger('i.suggest')
+
+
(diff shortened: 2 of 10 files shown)


willkg:
I'm interested in thoughts and comments. There isn't much test-wise, but there isn't much code, yet, either.

Thoughts?

fjord/suggest/utils.py
```diff
@@ -0,0 +1,20 @@
+import logging
+
+from fjord.suggest import get_providers
+
+
+logger = logging.getLogger('i.suggest')
+
+
+def get_suggestions(response):
+    """Given a feedback response, returns an ordered list of suggestions"""
```
willkg: I haven't figured out how to do "ordered" here. Right now it's ordered in the sense that the list of providers add suggestions in order of providers in `SUGGEST_PROVIDERS`. I think I might keep it that way, but it does create some scaling problems.

What happens if we get 20 suggestions back? We can't show them all on the Thank You page, so we'll have to trim the list. Do we want the 5 most important ones? If so, how do we order by importance? Etc.

Pushing most of that off for now since we're no where near that scenario, yet.

fjord/suggest/utils.py
```diff
@@ -0,0 +1,20 @@
...
+    suggestions = []
+
+    for prov in get_providers():
+        try:
+            suggestions.extend(prov.get_suggestions(response))
```
rlr: I think this is the first time I see extend. My learning for the day :+1:

fjord/suggest/apps.py
```diff
@@ -0,0 +1,19 @@
...
+from fjord.base.plugin_utils import load_providers
+
+
+logger = logging.getLogger('i.suggest')
+
+
+class SuggestConfig(AppConfig):
+    name = 'fjord.suggest'
+
+    def ready(self):
```
rlr: :+1: yay!

fjord/suggest/apps.py
```diff
@@ -0,0 +1,19 @@
...
+        # Load providers and store them in _PROVIDERS stomping on
+        # whatever was there.
+        _PROVIDERS[:] = load_providers(settings.SUGGEST_PROVIDERS, logger)
```
willkg: For the record, the `_PROVIDERS[:]` on the left-hand side means "replace the contents of `_PROVIDERS` with the value on the right hand side". Thus `_PROVIDERS` points to the same list, but we've changed the contents of the list. It's a cheap way of getting around labeling it as a "global" and then replacing it verbatim which has a bunch of unexciting consequences related to importing names and being difficult to mock and such.

rlr:
This looks good to me. I don't know if it will last 400 years like :notes: :musical_note: 

It will do for now though.

fjord/base/plugin_utils.py
```diff
@@ -0,0 +1,44 @@
...
+    :returns: list of providers.
+
+    >>> load_providers(['some.path.Class', 'some.other.path.OtherClass'])
+    [<Class>, <OtherClass>]
+
+    .. Note::
+
+       All exceptions when importing and loading providers are
+       swallowed and logged as errors. Providers that don't import or
+       don't load are nmot returned in the return list.
```
mythmon: Typo: nmot

willkg: Doh!

willkg:
^^^ rebases against master and fixes the typo.

willkg:
@mythmon points out it's probably better to do instances. I'll switch it now.

willkg:
^^^ Redoes it so it uses instances instead of classes.

rlr:
LGTM. :sunny:

willkg:
Thank you!

PR #565, opened 2015-05-04
[bug 1144159] Remove usage of django-celery.

rlr:
* Use celery the way suggested in the docs.
* Create a celeryd command so our infra keeps working without changes.
* Move some monkeypatching stuff around.

There are some things that need to be cleaned up. Wanted to posted it like this so we can discuss. I'll add some comments inline.

f?

### DIFF
--- a/fjord/__init__.py
+++ b/fjord/__init__.py
@@ -6,11 +6,6 @@ import os
 import site
 import sys
 
-from fjord.manage_utils import monkeypatch
-
-
-os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'fjord.settings')
-
 
 # Denotes that setup_vendor_path has been run.
 _has_setup_vendor_path = False
@@ -58,8 +53,6 @@ def setup_vendor_path():
 
 
 setup_vendor_path()
-
-monkeypatch()
 
 
 # Setup celery
--- a/fjord/__init__.py
+++ b/fjord/__init__.py
@@ -1,6 +1,15 @@
+# First we import absolute imports from the future, so that our celery.py
+# module will not clash with the library
+from __future__ import absolute_import
+
 import os
 import site
 import sys
+
+from fjord.manage_utils import monkeypatch
+
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'fjord.settings')
 
 
 # Denotes that setup_vendor_path has been run.
@@ -49,3 +58,11 @@ def setup_vendor_path():
 
 
 setup_vendor_path()
+
+monkeypatch()
+
+
+# Setup celery
+# This will make sure the app is always imported when
+# Django starts so that shared_task will use this app.
+from .celery import app as celery_app
--- a/fjord/base/management/commands/celeryd.py
+++ b/fjord/base/management/commands/celeryd.py
@@ -11,12 +11,12 @@ import sys
 from django.core.management.base import BaseCommand
 
 import celery
-from celery.bin import worker
+from celery.bin.worker import worker as WorkerCommand
 
 from fjord.celery import app
 
 
-worker = worker.worker(app=app)
+worker = WorkerCommand(app=app)
 
 
 # This code is based on django-celery code. It's a merge of
@@ -31,7 +31,6 @@ class Command(BaseCommand):
     requires_model_validation = False
     keep_base_opts = False
     stdout, stderr = sys.stdout, sys.stderr
-
 
     def handle(self, *args, **options):
         worker.check_args(args)
--- a/fjord/base/management/commands/celeryd.py
+++ b/fjord/base/management/commands/celeryd.py
@@ -1,0 +1,89 @@
+"""
+
+Start the celery daemon from the Django management command.
+
+"""
+from __future__ import absolute_import, unicode_literals
+
+import os
+import sys
+
+from django.core.management.base import BaseCommand
+
+import celery
+from celery.bin import worker
+
+from fjord.celery import app
+
+
+worker = worker.worker(app=app)
+
+
+# This code is based on django-celery code. It's a merge of
+# djcelery.management.base.CeleryCommand and
+# djcelery.management.commands.celeryd.Command.
+class Command(BaseCommand):
+    help = 'Old alias to the "celery worker" command.'
+    options = (BaseCommand.option_list
+               + worker.get_options()
+               + worker.preload_options)
+    skip_opts = ['--app', '--loader', '--config', '--no-color']
+    requires_model_validation = False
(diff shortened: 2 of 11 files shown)


fjord/__init__.py
```diff
@@ -1,7 +1,16 @@
+# First we import absolute imports from the future, so that our celery.py
+# module will not clash with the library
+from __future__ import absolute_import
+
 import os
 import site
 import sys
 
+from fjord.manage_utils import monkeypatch
```
rlr: ugh. so celery needs the monkeypatching to be in place too. Should we just move all the manage_utils stuff here?

fjord/__init__.py
```diff
@@ -1,7 +1,16 @@
...
+
+
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'fjord.settings')
```
rlr: this is probably in like 3 places now.

fjord/base/management/commands/celeryd.py
```diff
@@ -0,0 +1,89 @@
...
+
+from fjord.celery import app
+
+
+worker = worker.worker(app=app)
+
+
+# This code is based on django-celery code. It's a merge of
+# djcelery.management.base.CeleryCommand and
+# djcelery.management.commands.celeryd.Command.
```
rlr: It's possible some of the code in here below isn't needed.

fjord/celery.py
```diff
@@ -0,0 +1,22 @@
...
+
+# Using a string here means the worker will not have to
+# pickle the object when using Windows.
+app.config_from_object('django.conf:settings')
+app.autodiscover_tasks(lambda: settings.INSTALLED_APPS)
+
+
+@app.task(bind=True)
+def debug_task(self):
+    print('Request: {0!r}'.format(self.request))
```
rlr: this was cool to have for quick tests. Not sure if it's useful going forward.

fjord/__init__.py
rlr: To be clear, this is probably not needed if I do `./manage.py celeryd`. But if I do `celery -A fjord worker -l info` it is. And that's the future!

fjord/__init__.py
willkg: Only three? We should do better!

fjord/celery.py
```diff
@@ -0,0 +1,22 @@
+from __future__ import absolute_import
+
+import os
+
+from celery import Celery
+
+# set the default Django settings module for the 'celery' program.
+os.environ.setdefault('DJANGO_SETTINGS_MODULE', 'fjord.settings')
```
willkg: You probably don't need this one given that in order to import this module, you first have to import `fjord/__init__.py` which already has one of these.

fjord/celery.py
willkg: What's this?

fjord/manage_utils.py
```diff
@@ -63,6 +61,9 @@ def check_dependencies():
     from pip.req import parse_requirements
     from pip.download import PipSession
 
+    # Avoid circular import
+    from fjord import path
```
willkg: `path` is such a goofy thing. It pains me to have circular imports for something that's such a trivial and non-interesting bit of code. We should figure out what to do with `ROOT` and `path`.

fjord/base/management/commands/celeryd.py
```diff
@@ -0,0 +1,89 @@
...
+
+from django.core.management.base import BaseCommand
+
+import celery
+from celery.bin import worker
+
+from fjord.celery import app
+
+
+worker = worker.worker(app=app)
```
willkg: `worker` is a module, but then this line changes it into something else. Is there a better way to do this that doesn't involve overloading "worker"?

fjord/base/management/commands/celeryd.py
```diff
@@ -0,0 +1,89 @@
...
+    help = 'Old alias to the "celery worker" command.'
+    options = (BaseCommand.option_list
+               + worker.get_options()
+               + worker.preload_options)
+    skip_opts = ['--app', '--loader', '--config', '--no-color']
+    requires_model_validation = False
+    keep_base_opts = False
+    stdout, stderr = sys.stdout, sys.stderr
+
+
```
willkg: Nit: there should be only one blank line here.

fjord/__init__.py
```diff
@@ -49,3 +58,11 @@ def setup_vendor_path():
 
 
 setup_vendor_path()
+
+monkeypatch()
```
willkg: Why do we have to do the `monkeypatch()` call here? Which code path doesn't have monkeypatching?

fjord/celery.py
rlr: Oh, it's copy paste from the first step docs. I was using it to test celery from the shell.

```
In [1]: from fjord.celery import debug_task

In [2]: debug_task.delay()
Out[2]: <AsyncResult: 1ec26e79-ca12-49c4-b30c-e6bde4c49a5d>

In [3]: exit()
```

fjord/celery.py
rlr: ha! forgot about his one!

fjord/__init__.py
rlr: The code path of using the celery command as `$ celery -A fjord worker -l info` which is the suggested way now.

rlr: Note, that doesn't work in the vendor setup. I had to create a new virtualenv with requirements installed in it to test this path. I guess it should work in vagrant too.

fjord/__init__.py
willkg: I agree we want to switch to this, but not until after we ditch `vendor/`. So I vote we keep things simple for now and not do the monkeypatch thing here.

fjord/__init__.py
```diff
@@ -49,3 +58,11 @@ def setup_vendor_path():
...
+
+
+# Setup celery
+# This will make sure the app is always imported when
+# Django starts so that shared_task will use this app.
+from .celery import app as celery_app
```
willkg: I'm pretty sure we only need this for `celery -A ...` launching. Let's nix this for now.

willkg:
I created bug #1161236 to cover switching to "celery -A ...".

https://bugzilla.mozilla.org/show_bug.cgi?id=1161236

In the meantime, let's simplify this PR to just the bits we need to ditch djcelery and have our own celeryd Django command. That makes it easier to maintain for the next x months until we're in AWS land.

rlr:
I think that covers all the feedback so far :point_up:

willkg:
This looks good to me. Let's test it on -stage! r+

PR #567, opened 2015-05-06
[bug 1160495] Remove remaining south migrations.

rlr:
I think this is all that is left. No south in vendor or requirements. grepping for south and SOUTH only brings up something in a vendor lib.

r?


willkg:
Looks good to me. r+

When I wrote up the bug, I thought it was more involved, but looks like I had a bunch of empty directories.

PR #570, opened 2015-05-07
[bug 1133771] Rework thank you page view

willkg:
This reworks the Thank You page view and tests. We continue to use the
thankyou page flag, but this allows us to specify a response_id in the
querystring to make it easier to examine and debug behavior in
production.

Plus it gets rid of the thanks_sad.html template.

r?


fjord/feedback/views.py
```diff
@@ -44,23 +46,39 @@ def download_firefox(request, template):
...
-            opinion_id = request.session.get('opinion_id')
-            user_opinion = Response.objects.get(id=opinion_id)
-        except Response.ObjectDoesNotExist:
-            pass
-        else:
-            word_count = len(user_opinion.description.split())
-            if (user_opinion.locale == 'en-US' and
-                    not user_opinion.happy and
-                    word_count >= 7):
-                template = 'feedback/thanks_sad.html'
```
willkg: This code regarding sentiment, locale and word count check is getting moved to the SUMO Search Suggest API suggest provider.

fjord/feedback/tests/test_views.py
```diff
@@ -115,68 +117,60 @@ def test_valid_sad(self):
...
     @with_waffle('thankyou', True)
-    def test_use_thank_you_sad(self):
-        """Thanks_sad template is used when appropriate"""
+    def test_thankyou_flag_active(self):
+        """Verify response and suggestions when thankyou flag is active"""
         url = reverse('feedback', args=(u'firefox',), locale='en-US')
         r = self.client.post(url, {
             'happy': 0,
-            'description': u"I want to know why Firefox doesn't make me sandwiches!",
+            'description': u'Why Firefox not make me sandwiches!',
```
mythmon: This change was very important. :+1:

fjord/feedback/tests/test_views.py
```diff
@@ -115,68 +117,60 @@ def test_valid_sad(self):
...
         }, follow=True)
 
         feedback = models.Response.objects.latest(field_name='id')
-        eq_(u"I want to know why Firefox doesn't make me sandwiches!", feedback.description)
-        eq_(u'', feedback.url)
-        eq_(False, feedback.happy)
-        eq_(u'en-US', feedback.locale)
         eq_(r.status_code, 200)
-        self.assertTemplateUsed(r, 'feedback/thanks_sad.html')
+        eq_(r.context['response'].id, feedback.id)
```
mythmon: This is checking the context that was passed to render the template? I didn't know you could do that. Awesome!

fjord/feedback/tests/test_views.py
willkg: It's a cosmetic change, but the previous string was too long. We've still got Fjord on 79-char widths.

fjord/feedback/tests/test_views.py
```diff
@@ -115,68 +117,60 @@ def test_valid_sad(self):
...
+        eq_(r.context['suggestions'], None)
 
     @with_waffle('thankyou', True)
-    def test_use_default_thank_you_happy(self):
-        """Default thank you page is used when the feedback is happy."""
+    def test_thankyou_flag_and_response_id_in_qs_authenticated(self):
+        """Verify response_id in querystring overrides session id"""
+        # Create analyzer and log analyzer in
+        jane = ProfileFactory(user__email='jane@example.com').user
+        jane.groups.add(Group.objects.get(name='analyzers'))
```
mythmon: This code is fine, I don't think you need to change it. I'm curious though, could FactoryBoy make a `AnalyzerProfileFactory` that did this?

willkg: It totally could. That's a really good idea.

fjord/feedback/tests/test_views.py
```diff
@@ -115,68 +117,60 @@ def test_valid_sad(self):
...
+        self.client_login_user(jane)
+
+        feedback = ResponseFactory()
+        url = reverse('thanks') + '?response_id={0}'.format(feedback.id)
+        r = self.client.get(url)
```
mythmon: In the docstring, you say that the querystring should override the response id in the session. In this test though you don't have a response id in the session to override. Still a valid test, just not the one described in the docstring.

fjord/feedback/tests/test_views.py
```diff
@@ -115,68 +117,60 @@ def test_valid_sad(self):
...
+
+        eq_(r.status_code, 200)
+        eq_(r.context['response'].id, feedback.id)
+        eq_(r.context['suggestions'], [])
+
+    @with_waffle('thankyou', False)
+    def test_thankyou_flag_inactive(self):
+        """Verify response and suggestions when thankyou flag is active"""
```
mythmon: This docstring still says "active". It should say "inactive".

fjord/feedback/views.py
```diff
@@ -44,23 +46,39 @@ def download_firefox(request, template):
...
 
-    return render(request, template)
+    if waffle.flag_is_active(request, 'thankyou'):
+        response_id = None
+
+        # If the user is an analyzer/admin, then we let them specify
+        # the response_id via the querystring. This makes debugging
+        # the system easier.
+        if ((request.user.is_authenticated()
+             and request.user.has_perm('analytics.can_view_dashboard'))):
```
mythmon: I think you can skip the `request.user.is_authenticated()` part. The anonymous user won't have this permission. Or hey, maybe they will, because it was configured that way.

fjord/feedback/views.py
```diff
@@ -44,23 +46,39 @@ def download_firefox(request, template):
...
+            try:
+                response = Response.objects.get(id=response_id)
+            except Response.ObjectDoesNotExist:
+                pass
+            else:
+                suggestions = get_suggestions(response)
+
+    return render(request, 'feedback/thanks.html', {
+        'response': response,
+        'suggestions': suggestions
```
mythmon: There are no template changes in this PR that use `suggestions` (as far as I can see). Is that on purpose?

fjord/feedback/tests/test_views.py
mythmon: Oh. I thought it was just to make the sentence more like normal Input feedback. Now that I know there is a technical reason for it, I'm not as excited.

mythmon:
Ok, that's all I've got. Looks pretty good.

fjord/feedback/views.py
willkg: There's no suggestion providers, yet.

Basically, I've hit this point where there's a bunch of things I need to do each of which needs thought and they're sort of tangled:

1. build a SUMO Search Suggest API suggest provider
2. create mockups for the new thank you page

So I made these code changes to make it easier to work on those two things in tandem without having different versions of the thanks view.

fjord/feedback/tests/test_views.py
willkg: So, the impetus behind the change was the long line. However, I made the text like this because it's far more like real input and I thought it was kind of funny. I wanted to keep the general sentiment Adam had, too. It refers to the feedback description in another test in this file.

Another iteration I was toying with was "Y U NO MAKE SANDWICHES!!!!!" and then put enough ! to fill 79 characters. After seeing it in the code, I decided it was too horrible to force my future self or others to read.

fjord/feedback/tests/test_views.py
willkg: Created https://bugzilla.mozilla.org/show_bug.cgi?id=1162810 for that.

fjord/feedback/tests/test_views.py
willkg: Whoops! I deleted the code. Bah.

fjord/feedback/views.py
willkg: So this code is goof-tastic. It's semantically wrong, but I've been using it across the codebase. Now I just copy and paste it so that it's easier to replace it all when I get around to fixing it.

The problem is that this perm has nothing to do with being in the analyzer's group. I'm conflating having this permission with being part of the group, but I really should have a finer set of permissions that I should be testing which allow for multiple groups.

Anyhow, I'm going to keep it this way to make it easier to fix later.

willkg:
^^^ Fixes from feedback.

Does that look ok?

mythmon:
Looks good! r+

PR #576, opened 2015-05-14
[bug 1164910] Raise http 400 for bad arguments in alerts api

willkg:
This changes the alerts get api to raise an http 400 if it gets bad
arguments.

r?


mythmon:
Looks good.

willkg:
Awesome--thank you!

PR #577, opened 2015-05-15
[bug 1133769] Implement SUMO search suggest provider

willkg:
* move providers around so they're one folder higher--one less folder to
  deal with
* implement first pass of sumo search suggest provider

r?


fjord/suggest/provider_sumosuggest.py
```diff
@@ -0,0 +1,68 @@
...
+    params = {
+        'q': text,
+        'locale': 'en-US',
+        'product': 'firefox',
+        'max_documents': 3,
+        'max_questions': 0
+    }
+    headers = {
+        'content-type': 'application/json',
+        'accept': 'application/json; indent=4',
```
mythmon: You probably don't need the `indent=4` bit.

fjord/suggest/tests/test_sumosuggest.py
```diff
@@ -0,0 +1,157 @@
...
+    """
+    provider = SUMOSuggestProvider()
+
+    def test_get_suggestions(self):
+        resp = ResponseFactory(
+            happy=False,
+            locale=u'en-US',
+            product=u'Firefox',
+            description=(
+                u'slow browser please speed improve i am wait speed improve'
```
mythmon: it will be better

mythmon:
There is no error checking from the API, but you know that already. I don't really like getting rid of the providers directory. I think a bunch of files named `provider_foo.py` is really just making a directory without making one on the filesystem.

This looks like an ok first pass of the SUMO suggester, and certainly enough to start getting data from this. I don't see anything that is going to blow up.

fjord/suggest/provider_sumosuggest.py
willkg: It makes it easier when looking at raw HTTP request/response data, but you're right--I don't need it when not debugging. I'll nix it.

fjord/suggest/tests/test_sumosuggest.py
willkg: This is a real input feedback response.

willkg:
Bah. I forgot about the error checking. I'll work on that now.

fjord/suggest/tests/test_sumosuggest.py
mythmon: wat.

willkg:
^^^ That adds exception handling plus a test plus it fixes the timeout at 3.5 seconds which is probably too long.

fjord/suggest/provider_sumosuggest.py
```diff
@@ -0,0 +1,77 @@
...
+        SUMO_SUGGEST_API_URL,
+        data=json.dumps(params),
+        headers=headers,
+        # This is slightly larger than 3 which is what the requests
+        # docs suggest, but we really want the shortest timeout we can
+        # get away with since this happens during the http request
+        # handling cycle.
+        timeout=3.5
+    )
+    return resp.json()['documents']
```
mythmon: What happens when the request times out? I guess it will raise an exception, which will be caught in the provider?

willkg: It should raise a timeout related exception.

fjord/suggest/tests/test_sumosuggest.py
```diff
@@ -0,0 +1,189 @@
...
+    SUMOSuggestProvider
+)
+
+
+class SUMOSuggestProviderTestCase(TestCase):
+    provider = SUMOSuggestProvider()
+
+    def test_not_happy(self):
+        resp = ResponseFactory(
+            happy=True,  # Happy
```
mythmon: The naming of this test is inconsistent with the rest. Should it be `test_not_sad`?

mythmon:
This looks good. One little nit.

r+

willkg:
^^^ Fixes the nit. I need to land the other PRs before I want to land this one, so I'll let this sit for a bit.

willkg:
^^^ Rebase against master. I'll wait for Travis to thumbs-up and then mull over it some more.

I'm tossing around putting things back in the `providers/` directory. It's just so long. It'd be nice if we had a shorter synonym for "providers".

willkg:
^^^ I moved everything back into the `providers/` directory.

I'll wait for travis, then land it.

willkg:
Thank you!

PR #578, opened 2015-05-18
[bug 1161746] More alerts api code cleanup

willkg:
This cleans up the alerts api code a bit more:

* use DRF exceptions where appropriate
* move "aflavors" validation to serializer and make the error handling
  better so that it spits out all errors at once rather than just the
  first one it hit
* fix a typo in HTTP 400 for flavors
* add a NotFound exception--this exists in DRF3, so when we upgrade we
  can use theirs

r?

### DIFF
--- a/fjord/alerts/api_views.py
+++ b/fjord/alerts/api_views.py
@@ -7,7 +7,7 @@ import rest_framework.views
 
 from fjord.alerts.models import Alert, AlertFlavor, AlertSerializer, Link
 from fjord.api_auth.models import Token
-from fjord.base.api_utils import StrictArgumentsMixin
+from fjord.base.api_utils import NotFound, StrictArgumentsMixin
 
 
 class TokenAuthentication(authentication.BaseAuthentication):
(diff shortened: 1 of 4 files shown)


fjord/alerts/api_views.py
```diff
@@ -189,26 +189,29 @@ def post(self, request):
...
-                errors='Flavor not specified in payload.'
-            )
+            raise exceptions.ParseError({
+                'flavor': [
+                    'Flavor not specified in payload'
+                ]
+            })
 
         try:
             flavor = AlertFlavor.objects.get(slug=flavorslug)
```
mythmon: Are you duplicating the work to get and validate the flavor here? This seems the same as the work done in the validator. Plus, this version doesn't deal with comma separated values?

willkg: This is the POST API which takes a single flavor and doesn't currently use a serializer.

mythmon: Oh, oops. I misread the diff, and I thought this was still the get method.

willkg: No worries! The POST code should get rewritten, too, but I'll do that in a later pass and probably under a different "clean up all the things" bug.

mythmon:
Besides my mistake with the get/post views, this looks good. There is still more improvements that could be made, especially in the post API, but it probably belongs in another PR. This looks good to me.

willkg:
Awesome--thank you!

PR #579, opened 2015-05-18
1165996 description

willkg:
r?

### DIFF
--- a/fjord/feedback/migrations/0003_delete_empty_desc_feedback.py
+++ b/fjord/feedback/migrations/0003_delete_empty_desc_feedback.py
@@ -1,0 +1,36 @@
+# -*- coding: utf-8 -*-
+"""Deletes all feedback that have a description that's an empty
+string.
+
+"""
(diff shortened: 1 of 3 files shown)


willkg:
Pretty sure this is straight forward. I don't remember offhand why it was set for `blank=True` originally. Seems like it was a mistake.

mythmon:
Travis :exclamation:. Though looking at it, it is probably a transient failure. I'll restart it.

Wasn't there some source of feedback in the past that would only report sad/happy, but not give any description? That would explain why this was set to blank=True.

willkg:
I don't recall one. There are only 617 responses that have an empty string description in the db, so I suspect it's just a mistake.

willkg:
Travis passed. Yippie yay!

mythmon:
Yay, r+

willkg:
^^^ Rebased against master.

willkg:
Thank you!

PR #580, opened 2015-05-19
[bug 1132455] Upgrade to django-rest-framework 3.1.2

willkg:
* upgrade to django-rest-framework 3.1.2
* update and clean up some api code without any backwards incompatible
  changes

r?

### DIFF
--- a/fjord/alerts/api_views.py
+++ b/fjord/alerts/api_views.py
@@ -1,4 +1,3 @@
-from rest_framework import authentication
 from rest_framework import exceptions
 from rest_framework import permissions
 from rest_framework import serializers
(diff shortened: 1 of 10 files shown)


willkg:
This is kind of hard to review. It involves a lot of mediocre code getting converted in strange ways to new DRF3-compliant code. The tests are probably ok regarding making sure we didn't change the APIs in backwards-incompatible ways. I went through and did a lot of hand-testing for:

* posting feedback
* posting heartbeat data
* posting alerts
* getting alerts

Those are the public things that we're on the line for.

In doing this, I noticed a lot of API code that should get refactored using DRF3 things. DRF3 fixes a bunch of API things that'll make it easier to use those DRF3 things. In doing that, the API results will be more consistent from API to API. I wrote up bug #1166456 and friends to research that more and address those issues.

For now, I tried to minimize the change.

To review:

1. go through the code and the notes
2. run the tests
3. run the smoke tests

fjord/feedback/models.py
```diff
@@ -617,24 +617,11 @@ def __unicode__(self):
...
-    DRF lets CharFields take nulls which is not what I want. This
-    raises a ValidationError if the value is a null.
-
-    """
-    def from_native(self, value):
-        if value is None:
-            raise ValidationError('Value cannot be null')
-        return super(NoNullsCharField, self).from_native(value)
-
-
```
willkg: DRF3 distinguishes between the various possibilities of required, no-nulls, and no-empty-string so we can use that and don't need this `NoNullsCharField` anymore.

fjord/feedback/models.py
```diff
@@ -645,100 +632,103 @@ class PostResponseSerializer(serializers.Serializer):
...
+        # First, pop out 'email' since it's not in the Response.
+        validated_data.pop('email', None)
+
+        # Strip string data
+        for key in ['url', 'description', 'category', 'product',
+                    'channel', 'version', 'platform', 'locale',
+                    'manufacturer', 'device', 'country', 'source',
+                    'campaign', 'user_agent']:
+            if key in validated_data:
+                validated_data[key] = validated_data[key].strip()
```
willkg: I put this here because I wasn't sure where else to put it. Maybe it should be in the model?

Thoughts?

fjord/alerts/api_views.py
```diff
@@ -74,10 +28,9 @@ def is_after(value1, value2):
 
 
 class AlertsGETSerializer(StrictArgumentsMixin, serializers.Serializer):
+    """Serializer that validates GET API arguments"""
     flavors = serializers.CharField(required=True)
-    max = serializers.IntegerField(
-        required=False, default=100,
-        validators=[positive_integer])
+    max = serializers.IntegerField(default=100, min_value=1)
```
rlr: what happened with positive check? is that builtin now?

mythmon: I assume `min_value=1` takes care of that.

rlr: lol! #truth

fjord/feedback/models.py
mythmon: While working on APIs in SUMO for BuddyUp, I ran into a lot of cases where behavior was encoded only in the views, which produced a lot of annoying validation problems. Because of that, I am going to try and push things into the models in SUMO.

This seems appropriate for the models. Maybe in a save method? Or maybe these fields on the model should be a subclass of CharField that automatically strips?

fjord/alerts/api_views.py
willkg: Yeah! I figured `min_value=1` is less code than all the `positive_integer` stuff.

fjord/feedback/models.py
willkg: It could be a `CharField`. That's probably the best place for it since it's explicit and exactly where it needs to be.

rlr:
Changes LGTM. I'll go through both sets of tests next.

rlr:
yay got all smoketests to run and pass :fireworks:

willkg:
Thank you!

PR #581, opened 2015-05-20
[bug 1165996] Explicitly validate description in api

willkg:
What was happening was that if the description was whitespace, it would
get validated (it's not blank, so yippie!), and then get stripped and
saved. That resulted in blank descriptions.

This adds validation to make sure we don't get descriptions that are
whitespace.

r?

### DIFF
--- a/fjord/feedback/models.py
+++ b/fjord/feedback/models.py
@@ -1,7 +1,6 @@
 from datetime import datetime, timedelta
 
 from django.core.cache import cache
-from django.core.exceptions import ValidationError
 from django.db import models
 
 from elasticutils.contrib.django import Indexable, MLT
@@ -670,6 +669,13 @@ class PostResponseSerializer(serializers.Serializer):
         max_length=100, allow_null=False, default=u'')
     campaign = serializers.CharField(
         max_length=100, allow_null=False, default=u'')
+
+    def validate_description(self, value):
+        if not value.strip():
+            raise serializers.ValidationError(
+                u'This field may not be blank.'
+            )
+        return value
 
     def validate_url(self, value):
         if value:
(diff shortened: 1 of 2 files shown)


mythmon:
I understand how the serializer could let blank things through like you describe, and this change would prevent that. But shouldn't the model's validation reject empty strings since the `Response.description` field has the default value of `blank=True`?

rlr:
Tests pass for me. Travis is thinking...

rlr:
model validation only runs if you call a `model_instance.full_clean()` explicitly. `model_instance.save()` doesn't trigger validation (that would've been a big backwards incompatible change when model validation was introduced). so, `blank=True` only comes into play in form validation like in the admin, for example.

mythmon:
Oh. That is really surprising to me, and makes me want to add `self.full_clean()` to SUMO's `ModelBase.save()` method.

rlr:
Without giving it much thought, that sounds like a great idea.

willkg:
DRF has this blurb in the 3.0 announcement:

> This change also means that we no longer use the .full_clean() method on model instances, but instead perform all validation explicitly on the serializer. This gives a cleaner separation, and ensures that there's no automatic validation behavior on ModelSerializer classes that can't also be easily replicated on regular Serializer classes.

http://www.django-rest-framework.org/topics/3.0-announcement/#serializers

So it's possible it worked and then I broke it earlier today when I pushed out the DRF3 upgrade.

willkg:
Is `save()` too late in the process for validation work? At that point, the data should already be valid so drf and forms and views might not be expecting validation errors when saving.

mythmon:
Well, given all that, this change looks fine to me and looks like it fixes the problem.

Now I'm more thinking about SUMO's case where we have multiple views and APIs manipulating objects. Putting the validation anywhere but the model means duplicating it (or rewriting everything to always use serializers), which seems like a really bad idea to me. I can see how this could be beneficial for pure APIs, but it is a big step backwards for hybrid sites like SUMO. *grumble*.

willkg:
I'll wait for Travis to give the thumbs-up. Then land this. Thank you!

willkg:
Thank you!

PR #584, opened 2015-05-26
Time/timezone debugging view

willkg:
Add admin view for debugging possible time/timezone issues.

r?

### DIFF
--- a/fjord/base/admin.py
+++ b/fjord/base/admin.py
@@ -1,6 +1,7 @@
+from datetime import datetime
 import re
 import sys
-from datetime import datetime
+import time
 
 from django import VERSION
 from django.contrib import admin, messages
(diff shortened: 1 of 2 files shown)


willkg:
^^^ That fixes the Elasticsearch problem. We hand-index the item in Elasticsearch which nixes the possibility that celery takes too long to do things. Plus we add a 5-second sleep which helps to make sure all the bits are in place when we go to do the Elasticsearch query.

I feel pretty ok doing this given it's an admin page that's going to get used once in a blue moon. If this were to be used more often or was a user page, we wouldn't be doing these horrible things.

rlr:
WFM now. r+

willkg:
Thank you!

PR #585, opened 2015-05-26
[bug 1168475] Fix allow_blank on char fields

willkg:
* fix allow_blank=True on char fields that should allow blanks
* clean up some other code in what should be cosmetic ways

r?

### DIFF
--- a/fjord/feedback/models.py
+++ b/fjord/feedback/models.py
@@ -634,41 +634,41 @@ class PostResponseSerializer(serializers.Serializer):
     url = serializers.CharField(max_length=200, allow_blank=True, default=u'')
     description = serializers.CharField(required=True)
 
-    category = serializers.CharField(max_length=50, required=False,
-                                     default=u'')
+    category = serializers.CharField(
+        max_length=50, allow_blank=True, default=u'')
 
     # product, channel, version, locale, platform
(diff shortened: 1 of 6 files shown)


willkg:
This needs a test to verify the behavior. I'll write that tomorrow. Does this look right-ish, though?

fjord/feedback/models.py
```diff
@@ -634,41 +634,41 @@ class PostResponseSerializer(serializers.Serializer):
     url = serializers.CharField(max_length=200, allow_blank=True, default=u'')
     description = serializers.CharField(required=True)
 
-    category = serializers.CharField(max_length=50, required=False,
-                                     default=u'')
+    category = serializers.CharField(
+        max_length=50, allow_null=False, allow_blank=True, default=u'')
```
rlr: ah, you did what I was looking at. :+1:

rlr:
lgtm r+

willkg:
^^^ That nixes the `allow_null=False` because it's the default anyhow. It also adds a test for sending a blank category. Previously that would kick up an HTTP 400. Now we're ok with it.

rlr:
r+ if travis likes it!

willkg:
Thank you!

PR #586, opened 2015-05-27
[bug 1077596] Temporary histogram feedback API

willkg:
This is a temporary histogram feedback API. "Temporary" in this
case means we'll deprecate it as soon as we write a better one
using Elasticsearch aggregations. As soon as we do that, we'll
deprecate this one.

There's a bunch of FIXMEs in the code. We can fix them if this proves
less temporary than I'd like.

r?

### DIFF
--- a/fjord/feedback/api_views.py
+++ b/fjord/feedback/api_views.py
@@ -8,6 +8,7 @@ import rest_framework.response
 
 from elasticutils.contrib.django import F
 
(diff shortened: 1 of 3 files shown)


rlr:
I read the code. Looks normal.  I loaded the api view from the browser and it showed me some json.

good enough for me r+

willkg:
Thank you!

PR #590, opened 2015-06-03
[bug 1169261] Implement ga_track_event

willkg:
This adds a ga_track_event function that lets us capture data and post
it from the server. This is just like what we do currently with statsd,
but it's intended specifically for events and pushes the data to Google
where we can look at event flows.

r?

### DIFF
--- a/fjord/base/tests/test_google_utils.py
+++ b/fjord/base/tests/test_google_utils.py
@@ -1,0 +1,101 @@
+import sys
+from functools import wraps
+
+import requests_mock
+from mock import MagicMock, patch
+from nose.tools import eq_
+
+from fjord.base.google_utils import GOOGLE_API_URL, ga_track_event
+from fjord.base.tests import TestCase
+
+
+def set_sys_argv(sys_argv):
+    """Decorator for setting sys.argv
+
+    ::
+
+        class SomeTestCase(TestCase):
+            @set_sys_argv(['test'])
+            def test_something(self):
+                ...
+
+    """
+    def _set_sys_argv(fun):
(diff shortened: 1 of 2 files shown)


willkg:
@rlr Can you look at this when you get a chance? After this, I have one more redirector-related PR. Then I'm done with the redirector stuff.

fjord/base/tests/test_google_utils.py
```diff
@@ -0,0 +1,101 @@
+from functools import wraps
+import sys
+
+from mock import MagicMock, patch
+from nose.tools import eq_
+import requests_mock
```
rlr: I thought all the `import xxx` were supposed to go above the `from xxx import xxx`. /me looks up webdev bootcamp

rlr: `Imports on top, from-imports below:` http://mozweb.readthedocs.org/en/latest/reference/python-style.html#import-statements

rlr:
I just have that nit. Otherwise this is cool. I'm looking forward to see how it's used.

fjord/base/tests/test_google_utils.py
willkg: Yeah... I waffle between that and putting them in order alphabetically. At one point, we were throwing together a linter for it, but then I decided I just don't care a whole lot.

But I'll fix it and write it down in the docs so I remember in the future.

willkg:
^^^ Reorders imports. I'll wait for travis and then land it.

willkg:
Thank you!

PR #591, opened 2015-06-03
[bug 1169261] Implement redirector

willkg:
* implement redirector app
* add google_utils for pinging the server with events
* wrote SUMO Suggest redirector
* rework SUMO Suggester to use the redirector
* deal with the AAQ link
* minor renaming of things because I like to rename things over and over
  again when it's raining

r?

### DIFF
--- a/fjord/redirector/base.py
+++ b/fjord/redirector/base.py
@@ -1,0 +1,21 @@
+import urllib
+
+from fjord.base.urlresolvers import reverse
+
+
+class Redirector(object):
+    def load(self):
+        """Implement this to perform any startup work"""
+        pass
+
+    def handle_redirect(self, request, redirect):
+        """Implement this to handle redirects for a given request
+
+        :returns: None or url to redirect to
+
+        """
+        raise NotImplementedError
+
+
+def build_redirect_url(redirect):
+    return reverse('redirect-view') + '?' + urllib.urlencode({'r': redirect})
--- a/fjord/suggest/providers/sumosuggest.py
+++ b/fjord/suggest/providers/sumosuggest.py
@@ -2,8 +2,11 @@ import json
 import logging
 
 import requests
+from tower import ugettext as _
 
+from fjord.base.google_utils import ga_track_event
 from fjord.suggest import Link, Suggester
+from fjord.redirector import Redirector, build_redirect_url
 
 
 PROVIDER = 'sumosuggest'
@@ -11,7 +14,12 @@ PROVIDER = 'sumosuggest'
 # suggestions.
 PROVIDER_VERSION = 1
 
-SUMO_SUGGEST_API_URL = 'https://support.mozilla.org/api/2/search/suggest/'
+EVENT_CATEGORY = 'sumo_suggest'
+
+SUMO_HOST = 'https://support.mozilla.org'
+SUMO_SUGGEST_API_URL = SUMO_HOST + '/api/2/search/suggest/'
+SUMO_SUGGEST_SESSION_KEY = 'sumo_suggest_docs_{0}'
+SUMO_AAQ_URL = 'https://support.mozilla.org/questions/new'
 
 
 logger = logging.getLogger('i.sumosuggest')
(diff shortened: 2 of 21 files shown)


willkg:
The redirector is an app that manages the `/redirect` view. It takes an `r` parameter which it then passes through a list of redirector providers as specified in `settings.REDIRECTOR_PROVIDERS`. The first one to return a url wins and then the user is redirected to that url.

This creates a SUMO Suggest redirector. This encodes the rank of the url to redirect to or "aaq" in the redirection value. The rank is the index of the document stored in the user's session which was returned from the SUMO Suggest API.

In this way:

1. no one can see anyone else's redirections
2. redirection can't (currently) be abused to send users to "naughty places"
3. redirections and documents expire with the django session
4. we capture the appropriate information and send it to google so we can see event flows and compare "# suggests" to "# number of aaq links pressed" and "# non-aaq links pressed" and use this to tune the suggestion algorithm and/or even know whether it's useful at all

That's the gist of it.

fjord/redirector/base.py
```diff
@@ -0,0 +1,21 @@
...
+        """Implement this to handle redirects for a given request
+
+        :returns: None or url to redirect to
+
+        """
+        raise NotImplementedError
+
+
+def build_redirect_url(redirect):
+    return reverse('redirect-view') + '?' + urllib.urlencode({'r': redirect})
```
rlr: should this use `fjord.base.helpers.urlparams()` instead?

fjord/suggest/providers/sumosuggest.py
```diff
@@ -46,11 +54,115 @@ def get_kb_articles(locale, product, text):
...
+        # Extract the rank.
+        try:
+            trigger, rank = parse_redirect(redirect)
+        except RedirectParseError:
+            return
+
+        if rank == 'aaq':
+            url = SUMO_AAQ_URL
+        else:
+            # This doc doesn't exist.
```
rlr: this comment is confusing. does it go inside the `except` below?

rlr:
I wasn't expecting the session to be used, that's awesome for all the reasons you noted. I'm going to chew on that for a little but I think I am :+1: to that.

fjord/redirector/base.py
willkg: Mmm... `urlparams` is a template filter and it does a bunch of other stuff like handling duplicates that we don't need. I think I'm on the fence.

fjord/suggest/providers/sumosuggest.py
willkg: Whoops--you're right.

willkg:
The other ways to do it:

1. store suggestions in the db which requires us to do some more implementation and also write some code to clean them up later
2. store suggestions in cache, but i'm concerned that it'll get bumped out of cache which would cause a bunch of problems with the redirects

Are there other options? Seems like for something like this where it's a short list of dicts, putting it in the session works.

willkg:
I was waiting for the microwave to reheat my frittata and mulling over everything when it occurred to me that if you leave two feedbacks with the same django session, then the sumo suggest provider will show you suggestions from the first feedback because it's not storing the response id. Pretty sure that's true. I'll look into it now.

Also interesting, the spelling-correct thing wants to replace "frittata" with "fritter".

willkg:
^^^ That fixes the "comment in the wrong place" problem plus my "sumosuggest provider should be cleverer about docs in the session" problem.

rlr:
I like. Gonna run tests since travis doesn't like us today

fjord/suggest/tests/test_sumosuggest.py
```diff
@@ -97,86 +157,113 @@ def test_suggestions(self):
         }
 
         with requests_mock.Mocker() as m:
-            m.get(SUMO_SUGGEST_API_URL, json=ret)
+            m.get(SUMO_SUGGEST_API_URL, json=sumo_api_ret)
+            m.post(GOOGLE_API_URL, text='whatever')
```
rlr: I'm digging the requests mock. Very cool

rlr:
I ran tests and read the code again. r+

willkg:
Awesome--thank you!

willkg:
^^^ Rebased.

PR #593, opened 2015-06-03
[bug 1169261] Fix event tracking

willkg:
Event tracking needs the provider version number so we can distinguish
between flows for successive algorithms. Further, the "suggest" event
needs to be synchronous while the "view*" events can be asynchronous.

Sorry--forgot to do this. :(

r?


rlr:
lgtm r+!

willkg:
Thank you!

PR #594, opened 2015-06-09
1133771 typ

willkg:
OMG!

r?


willkg:
To test:

1. run the tests
2. do `./manage.py runserver`
3. log in as an admin
4. leave some sad feedback for Firefox with a long description like "omg! video never works and all i see are black boxes filled with @rehandalal's head laughing at me!"
5. make sure the thank you page looks ok
6. add "thankyou=1" to the end of the thanks url you end up at
7. make sure the enhanced thank you page with suggestions looks ok
8. test the redirects
9. (optional) check google analytics and see if you can see evidence of the flow you just did though you might not be able to because awesome but know that i've tested this aspect already and feel ok about it and will push out hotfixes if it's not working later which would be ok

rlr:
Tests pass. I can't get the thing to work locally. Not sure if you forgot a step between 3 and 4 to allow the flag to be tested. But I applied the hammer and activated the flag for Everyone. Still nothing.

willkg:
Adding "thankyou=1" to the querystring should activate the flag. Is your `fjord/settings/local.py` overriding `SUGGEST_PROVIDERS` and `REDIRECTOR_PROVIDERS`?

willkg:
^^^ That alphabetizes the css things. Otherwise it's identical to before.

rlr:
I added this print statement:

```
    # If we don't have a response_id, then pull it from the
    # session where it was placed if the user had just left
    # feedback.
    if response_id is None:
        response_id = request.session.get('response_id')

    print response_id
```

It always prints 0.

rlr:
I see the query string trick, doing it that way does work (http://localhost:8000/en-US/thanks?response_id=3616)

willkg:
^^^ That should fix the `response_id` for admin issue.

rlr:
looks awesome! r+

willkg:
^^^ Quick rebase. I'll wait for Travis and then land it.

Thank you!

PR #596, opened 2015-06-10
[bug 1167192] In python, stop using the gettext from tower.

rlr:
As mentioned in IRC, I created fjord/base/translation.py to create the lazy versions. I am open to other ideas.

r?

### DIFF
--- a/docs/conventions.rst
+++ b/docs/conventions.rst
@@ -101,7 +101,7 @@ templates.
 
 In Python::
 
-    from tower import ugettext as _, ugettext_lazy as _lazy
+    from django.utils.translation import ugettext as _, ugettext_lazy as _lazy
 
     yodawg = _lazy('The Internet')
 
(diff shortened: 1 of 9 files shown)


fjord/base/translation.py
```diff
@@ -0,0 +1,6 @@
+from django.utils.functional import lazy
+from django.utils.translation import ugettext, ungettext
+
+
+ugettext_lazy = lazy(ugettext, unicode)
+ungettext_lazy = lazy(ungettext, unicode)
```
willkg: It looks like django has lazy versions already:

https://docs.djangoproject.com/en/1.8/topics/i18n/translation/#lazy-translation

Does that not work for what we're doing?

rlr: oooh. thanks for looking into that! I'll use that

rlr:
fixord ^^

willkg:
I ran the tests and they're fine. (No surprise there.)

I went through the feedback in French and didn't see any missing strings.

That's probably good enough for this. r+

PR #598, opened 2015-06-11
[bug 1172019] Make thank you page product-agnostic

willkg:
This creates a copy of the thank you page which has strings that are
product agnostic. As soon as the strings are translated, we'll ditch the
old version for this version.

r?

### DIFF
--- a/fjord/feedback/templates/feedback/thanks_l10nized.html
+++ b/fjord/feedback/templates/feedback/thanks_l10nized.html
@@ -1,0 +1,98 @@
+{% extends "feedback/base.html" %}
+
+{% block page_title %}{{ _('Thanks') }}{% endblock %}
+
+{% block site_js %}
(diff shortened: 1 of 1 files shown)


willkg:
This isn't used by anything, so the strings will get extracted, but otherwise it doesn't affect anything at all.

To test:

1. copy `fjord/feedback/templates/feedback/thanks_l10nized.html` to `fjord/feedback/templates/feedback/thanks.html`
2. run tests
3. run `./manage.py runserver`
4. go to `/thanks` and make sure it looks ok--it'll use Firefox as the default (this is the "no product specified" case)
5. leave feedback for Firefox for Android and make sure the thank you page makes sense

rlr:
I did all that and also did a diff after copying the file... All makes sense. r+

willkg:
Thank you!

PR #600, opened 2015-06-15
[bug 1173763] Log everything

willkg:
This logs everything to the journal so we can figure out what the
problem is.

Note: This is temporary. We'll remove this as soon as we figure it out.

Quick r?


rlr:
ship it! r+

willkg:
Thank you!

PR #601, opened 2015-06-15
[bug 1167192] Changed text domain for L10n from messages to django.

rlr:
To test this, I....

1- Did an extract+merge and verified django.pot and all the django.po files were created.
2- I looked at the django.po files and they didn't have translations, as expected.
3- I copied locale/es/LC_MESSAGES/messages.po to locale/es/LC_MESSAGES/django.po
4- Compiled locales and ran the server.
5- Verified that /es/ is localized but /fr/, for example, isn't

I think that proves that all the parts are right. What did I miss?

r?

### DIFF
--- a/bin/test_locales.sh
+++ b/bin/test_locales.sh
@@ -22,9 +22,9 @@ echo "extract and merge...."
 echo "creating dir...."
 mkdir -p locale/xx/LC_MESSAGES
 
-echo "copying messages.pot file...."
-cp locale/templates/LC_MESSAGES/messages.pot locale/xx/LC_MESSAGES/messages.po
+echo "copying django.pot file...."
+cp locale/templates/LC_MESSAGES/django.pot locale/xx/LC_MESSAGES/django.po
 
-echo "translate messages.po file...."
-./manage.py translate --pipeline=html,pirate locale/xx/LC_MESSAGES/messages.po
+echo "translate django.po file...."
+./manage.py translate --pipeline=html,pirate locale/xx/LC_MESSAGES/django.po
 locale/compile-mo.sh locale/xx/
--- a/fjord/settings/base.py
+++ b/fjord/settings/base.py
@@ -63,7 +63,7 @@ USE_I18N = True
 USE_L10N = True
 
 # Gettext text domain
-TEXT_DOMAIN = 'messages'
+TEXT_DOMAIN = 'django'
 STANDALONE_DOMAINS = [TEXT_DOMAIN]
 TOWER_KEYWORDS = {'_lazy': None}
 TOWER_ADD_HEADERS = True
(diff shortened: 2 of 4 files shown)


bin/test_locales.sh
```diff
@@ -23,8 +23,8 @@ echo "creating dir...."
 mkdir -p locale/xx/LC_MESSAGES
 
 echo "copying messages.pot file...."
-cp locale/templates/LC_MESSAGES/messages.pot locale/xx/LC_MESSAGES/messages.po
+cp locale/templates/LC_MESSAGES/django.pot locale/xx/LC_MESSAGES/django.po
 
 echo "translate messages.po file...."
```
willkg: Total nit, but we should update the comments here and in the previous comment, too.

fjord/settings/base.py
```diff
@@ -63,8 +63,8 @@ def path(*dirs):
 USE_L10N = True
 
 # Gettext text domain
-TEXT_DOMAIN = 'messages'
-STANDALONE_DOMAINS = [TEXT_DOMAIN, 'javascript']
+TEXT_DOMAIN = 'django'
+STANDALONE_DOMAINS = [TEXT_DOMAIN, 'djangojs']
```
willkg: I just pushed a change to this nixing the 'javascript' domain because we don't have any files and doing `./manage.py merge` was kicking up errors.

willkg:
Everything here looks ok to me. I don't see other instances of "messages.po" anywhere. r+

rlr:
fixed those two comments ^^

fjord/settings/base.py
willkg: We don't need `djangojs` here. It's a domain we don't use.

willkg:
I think if you fix the `djangojs` thing, this will be mergeable again.

rlr:
removed the djangojs thing ^^

fjord/settings/base.py
```diff
@@ -63,8 +63,8 @@ def path(*dirs):
 USE_L10N = True
 
 # Gettext text domain
-TEXT_DOMAIN = 'messages'
-STANDALONE_DOMAINS = [TEXT_DOMAIN, 'javascript']
+TEXT_DOMAIN = 'django'
+STANDALONE_DOMAINS = [TEXT_DOMAIN]
```
willkg: Instead of making this change, rebase against master.

rlr: gah! I see what you were saying

rlr:
yay. looks mergeable now

willkg:
Sweet--this looks super! So we just need to land this when we do the svn copy thing.

willkg:
The svn stuff is done. I'll run this through its paces and then land it and then push it to stage where we can double-check things are ok.

PR #603, opened 2015-06-16
[bug 1173763] Remove logging from sumosuggest provider

willkg:
For the last day, we logged all redirection urls in an attempt to see
whether Input was sending bad data to GA. After looking over 70+
entries, I see no evidence Input is sending bad data to GA.

Given that, I can only assume that Input is fine and GA is displaying
it incorrectly in the Event Flow report.

Given that, I'm removing the logging code.

Quick r?


rlr:
r+ !

willkg:
Thank you!

PR #604, opened 2015-06-17
[bug 1167192] Replace tower's activate().

rlr:
From the django docs, I'm pretty sure this is how you activate a locale. https://docs.djangoproject.com/en/1.8/ref/utils/#django.utils.translation.activate

Tower does a bunch of other stuff that I thought we didn't need. But this doesn't work. What am I missing?

### DIFF
--- a/fjord/base/l10n.py
+++ b/fjord/base/l10n.py
@@ -1,0 +1,13 @@
+from django.utils.translation import ugettext, ungettext
+
+
+def install_gettext():
+    """Install gettext into the Jinja2 environment."""
+    class Translation(object):
+        # We pass this object to jinja so it can use django's gettext
+        # implementation.
+        ugettext = staticmethod(ugettext)
+        ungettext = staticmethod(ungettext)
+
+    import jingo
+    jingo.env.install_gettext_translations(Translation)
--- a/fjord/base/middleware.py
+++ b/fjord/base/middleware.py
@@ -3,13 +3,13 @@ from warnings import warn
 
 from django.conf import settings
 from django.http import HttpResponsePermanentRedirect
+from django.utils import translation
 from django.utils.encoding import smart_str
-
-import tower
 
 from fjord.base import urlresolvers
 from fjord.base.browsers import parse_ua
 from fjord.base.helpers import urlparams
+from fjord.base.l10n import install_gettext
 
 
 """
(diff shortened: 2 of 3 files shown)


rlr:
For reference, django has this view for setting locale:
https://github.com/django/django/blob/1.7.8/django/views/i18n.py#L18

But that just confused me more now.

willkg:
Talked to @rlr about this on IRC:

1. pretty sure if we fix `fjord/settings/base.py` `LOCALE_PATHS`, then this works--it's currently looking at the wrong directory for `.mo` files.
2. we might have to call `translation.to_locale(prefixer.locale)` to get the right code. i think we can test this with `pt-BR`. if it works, then we're fine.

rlr:
I tested `pt-BR` and it works fine without the `to_locale` call. According to the docs, `activate` takes a language and not a locale. So I guess we are fine.

fjord/base/l10n.py
```diff
@@ -0,0 +1,61 @@
...
+
+    message = add_context(context, stripped) if context else stripped
+
+    ret = django_ugettext(message)
+
+    # If the context isn't found, we need to return the string without it
+    return stripped if ret == message else ret
+
+
+def ungettext(singular, plural, number, context=None):
```
rlr: can we just use django's gettext like in python?

rlr:
Oops, I tested pt-BR on master branch. I'll test for realz now.

rlr:
^^ I removed those gettext implementations and used django's instead. And fixed the `LOCALE_PATHS` setting. now ready for r?

willkg:
Wow--scarily short. I'll test it out now.

fjord/base/middleware.py
```diff
@@ -163,4 +163,5 @@ def process_request(self, request):
 
         request.path_info = '/' + prefixer.shortened_path
         request.locale = prefixer.locale
-        tower.activate(prefixer.locale)
+        translation.activate(prefixer.locale)
+        install_gettext()
```
willkg: Any idea why we need this line or why we have to do this every time we process a request?

rlr: I do know that without it, templates blow up because they don't have gettext. This will change with django-jinja though so it is kind of temporary.

rlr: Oh, your point is more about every request. I see... Maybe this should go into an `AppConfig.ready(self):...` ?

willkg: That sounds like it'd work.

I'm ok with it here, too. It's pretty menial. And if it goes away with django-jinja, then maybe it doesn't make sense to fiddle with things.

willkg:
This looks good to me as far as I can tell. r+!

PR #605, opened 2015-06-17
Fix to use better python

willkg:
We no longer use Python 2.6, so the default here should be whatever the
system python is which is 2.7 on my machine.

Quick r?


willkg:
The deploy script uses the `-p` flag, so it specifies the python explicitly. This won't affect deploys.

rlr:
r+

PR #607, opened 2015-06-17
[bug 1167192] Copy Jinja2 i18n extension from tower.

rlr:
I only brought the bits that collapse whitespace in blocks.

r?

### DIFF
--- a/fjord/base/l10n.py
+++ b/fjord/base/l10n.py
@@ -1,4 +1,9 @@
+import re
+
 from django.utils.translation import ugettext, ungettext
+
+import jinja2
+from jinja2.ext import InternationalizationExtension
 
 
 def install_gettext():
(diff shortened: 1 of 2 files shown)


willkg:
This looks awesome. I want to take it through its paces. I'll do that tonight.

fjord/base/l10n.py
```diff
@@ -11,3 +16,19 @@ class Translation(object):
...
+class MozInternationalizationExtension(InternationalizationExtension):
+    """
+    We override jinja2's _parse_block() to collapse whitespace so we can have
+    linebreaks wherever we want.
+    """
+
+    def _parse_block(self, parser, allow_pluralize):
+        parse_block = InternationalizationExtension._parse_block
+        ref, buffer = parse_block(self, parser, allow_pluralize)
+        return ref, strip_whitespace(buffer)
```
willkg: Is this better written as:

```
def _parse_block(self, *args, **kwargs):
    ref, buff = super(MozInternationalizationExtension, self)._parse_block(*args, **kwargs)
    return ref, strip_whitespace(buff)
```

?

willkg:
I had that one nit. Otherwise this looks ok. I'll test it out now.

fjord/base/l10n.py
rlr: That's does look like the right way... I'll change it and make sure stuff still works

willkg:
Looks good testing-wise. I think I'll just land it and we can deal with the nit later if we feel like it.

rlr:
thank you!

PR #608, opened 2015-06-18
[bug 1167192] Copy tower commands over and remove tower dependency.

rlr:
r?

### DIFF
--- a/fjord/base/l10n.py
+++ b/fjord/base/l10n.py
@@ -1,8 +1,11 @@
+# NOTE: The code in this file was mostly copied from tower.
 import re
 
 from django.utils.translation import ugettext, ungettext
 
 import jinja2
+from babel.messages.extract import extract_python
+from jinja2 import ext
 from jinja2.ext import InternationalizationExtension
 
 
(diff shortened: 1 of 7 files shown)


willkg:
We both noticed that doing an extract/merge with these code changes redoes the wrapping in the uk django.po file as well as a few other locales. It doesn't change anything interesting in the django.pot file and some locales had no changes at ll.

I switched back to the master branch and did a fresh extract/merge and saw the same changes in the uk django.po file. Ergo, I think it's just rewrapping that .po file from whatever the translator committed and unrelated to these changes.

Figured I'd mention that on record.

fjord/base/l10n.py
```diff
@@ -32,3 +47,52 @@ def _parse_block(self, parser, allow_pluralize):
...
+                singular, plural, num = message
+                message = (strip_whitespace(singular),
+                           strip_whitespace(plural),
+                           num)
+        elif len(message) == 4:
+            singular, plural, num, ctxt = message
+            message = (add_context(ctxt, strip_whitespace(singular)),
+                       add_context(ctxt, strip_whitespace(plural)),
+                       num)
+    return message
```
willkg: We did a check earlier and there are no instances of msgctxt, so we could nix the msgctxt stuff here. However, it's probably better to keep the transition simple for now so I think we should just leave it as is.

willkg:
This looks good to me. r+

PR #611, opened 2015-06-29
[bug 1178440] Fix TEST_RUNNER warning

willkg:
We were getting a TEST_RUNNER warning which was spurious because we're
not using the Django default test runner. However, the warning kicked
off because there were some default settings in our settings.py file
which caused the checker to think we had upgraded from Django 1.5 and
thus needed the warning.

This removes some of those settings and tweaks the code so the spurious
warning doesn't pop up anymore.

r?

### DIFF
--- a/fjord/settings/base.py
+++ b/fjord/settings/base.py
@@ -9,13 +9,8 @@ from django.utils.functional import lazy
 
 from fjord.settings_utils import config
 
-
-ROOT = os.path.abspath(
-    os.path.join(
-        os.path.dirname(__file__),
-        '..',
-        '..'
-    ))
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+ROOT = os.path.dirname(BASE_DIR)
 
 
 def path(*dirs):
@@ -29,7 +24,6 @@ PROJECT_MODULE = 'fjord'
 ROOT_URLCONF = '%s.urls' % PROJECT_MODULE
 
 ADMINS = ()
-MANAGERS = ADMINS
 DEV = False
 
 DATABASES = {'default': config('DATABASE_URL', type_='database_url')}
@@ -39,9 +33,6 @@ SLAVE_DATABASES = []
 DATABASE_ROUTERS = ('multidb.PinningMasterSlaveRouter',)
 
 CACHES = {'default': config('CACHE_URL', type_='cache_url')}
-
-# Site ID is used by Django's Sites framework.
-SITE_ID = 1
 
 ## Internationalization.
 


willkg:
Tests should run fine.

Running `./manage.py` from master shows the warning. If you run `./manage.py` with this branch, there is no warning.

The check in question is here:

https://github.com/django/django/blob/8830c05c262cc9f5cbfca2544eaf083726551113/django/core/checks/compatibility/django_1_6_0.py#L17

The old code ended up with a high weight. The new code it's 4 which doesn't trigger the spurious warning.

fjord/settings/base.py
```diff
@@ -9,13 +9,8 @@
...
 from fjord.settings_utils import config
 
-
-ROOT = os.path.abspath(
-    os.path.join(
-        os.path.dirname(__file__),
-        '..',
-        '..'
-    ))
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
```
willkg: Django 1.6 introduced `BASE_DIR`. I added it, but we only use it on the next line. I figured it might make future things easier since it's a standard setting in the Django template.

fjord/settings/base.py
```diff
@@ -29,7 +24,6 @@ def path(*dirs):
 ROOT_URLCONF = '%s.urls' % PROJECT_MODULE
 
 ADMINS = ()
-MANAGERS = ADMINS
```
willkg: `MANAGERS` is only used by the `mail_managers` utility which we don't use. Thus there's no need to set this.

fjord/settings/base.py
```diff
@@ -40,9 +34,6 @@ def path(*dirs):
 
 CACHES = {'default': config('CACHE_URL', type_='cache_url')}
 
-# Site ID is used by Django's Sites framework.
-SITE_ID = 1
```
willkg: We don't use the sites app, so we don't need this line.

fjord/settings/base.py
```diff
@@ -9,13 +9,8 @@
...
 
-
-ROOT = os.path.abspath(
-    os.path.join(
-        os.path.dirname(__file__),
-        '..',
-        '..'
-    ))
+BASE_DIR = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+ROOT = os.path.dirname(BASE_DIR)
```
willkg: `BASE_DIR` ends up on my machine as `/home/willkg/mozilla/fjord/fjord` -- the directory with the fjord apps in it.

`ROOT` ends up on my machine as `/home/willkg/mozilla/fjord` -- the git repository root.

rlr:
lgtm! r+

willkg:
Thank you!

PR #613, opened 2015-06-30
[bug 1177451] Return safe (jinja2.Markup) strings from _()

rlr:
r?

I admit my testing was lazy. Should I do something different?

### DIFF
--- a/fjord/base/l10n.py
+++ b/fjord/base/l10n.py
@@ -37,16 +37,28 @@ def strip_whitespace(message):
     return re.compile(r'\s+', re.UNICODE).sub(' ', message).strip()
 
 
+@jinja2.contextfunction
+def _gettext_alias(context, string, *args, **kw):
+    """Takes the result of gettext and marks it safe."""
+    return jinja2.Markup(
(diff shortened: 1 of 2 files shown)


fjord/analytics/tests/test_views.py
```diff
@@ -58,6 +58,9 @@ def test_front_page(self):
         eq_(pq('.block.count strong').text(), '7')
         eq_(len(pq('li.opinion')), 7)
 
+        assert '&amp;raquo;' not in r.content
+        assert '&raquo;' in r.content
```
willkg: Mmm... I'd put this in a separate test since it's making sure our gettext is set up right. Putting it here with no comments about it will be confusing when it fails.

willkg: Plus I'd put this test in `fjord/base/tests/` since it tests the l10n stuff which lives there.

willkg:
Other than my pedantic issues with the test, this looks good to me.

rlr:
fixord ^^

willkg:
AWESOME!

PR #614, opened 2015-07-13
[bug 1183113] Upgrade django to 1.7.9

willkg:
r?


willkg:
Travis passed it and no one is around, so I'm going to self-r+ this.

PR #615, opened 2015-07-13
[bug 1183130] Fix test_picker.py smoketest

willkg:
A few weeks ago, I changed the class of the thing that the test looked
at, but didn't update the test. This updates the test.

r?


willkg:
This is a fix for the smoketests. I think I'm the only person to use them, so I'm going to self-r+ this one.

PR #616, opened 2015-07-14
[bug 1183764] Add country field to hb answer model

willkg:
r?


willkg:
Travis is cool with it. This is a 2 line change plus an automatically generated migration. Plus I removed an import that had no effect on things.

I'm going to self-r+ this.

PR #617, opened 2015-07-15
[bug 1184273] Add country to hb api, tests and docs

willkg:
r?


willkg:
There are two lines of code added to the tests. Otherwise, this is mostly a docs change.

Given that, I'm going to self-r+ it.

PR #618, opened 2015-07-16
[bug 1170333] es

willkg:
r?

### DIFF
--- a/fjord/analytics/analyzer_views.py
+++ b/fjord/analytics/analyzer_views.py
@@ -14,11 +14,11 @@
 # Also, since it's just analyzers, everyone is expected to speak
 # English. Ergo, there's no localization here.
 
-from datetime import datetime, timedelta
+from datetime import date, datetime, timedelta
 from pprint import pformat
 import csv
 
-from elasticutils.contrib.django import F, es_required_or_50x
+from elasticsearch_dsl import F
 
 from django.core.paginator import Paginator, PageNotAnInteger, EmptyPage
 from django.core.urlresolvers import reverse_lazy
@@ -32,7 +32,7 @@ from fjord.analytics.forms import (
     ProductsUpdateForm,
     SurveyCreateForm
 )
-from fjord.analytics.utils import counts_to_options
+from fjord.analytics.utils import counts_to_options, zero_fill
 from fjord.base.helpers import locale_name
 from fjord.base.utils import (
     analyzer_required,
@@ -41,9 +41,10 @@ from fjord.base.utils import (
     smart_date
 )
 from fjord.feedback.helpers import country_name
-from fjord.feedback.models import Product, Response, ResponseMappingType
+from fjord.feedback.models import Product, Response, ResponseDocType
 from fjord.heartbeat.models import Answer, Survey
 from fjord.journal.models import Record
+from fjord.search.index import es_required_or_50x
 from fjord.search.utils import es_error_statsd
 
 
@@ -166,12 +167,12 @@ def _analytics_search_export(request, opinions_s):
     keys = Response.get_export_keys(confidential=True)
     total_opinions = opinions_s.count()
 
-    opinions_s = opinions_s.values_list('id')[:MAX_OPINIONS]
+    opinions_s = opinions_s.fields('id')[:MAX_OPINIONS].execute()
 
     # We convert what we get back from ES to what's in the db so we
     # can get all the information.
     opinions = Response.objects.filter(
-        id__in=[mem[0][0] for mem in opinions_s])
+        id__in=[mem['id'][0] for mem in opinions_s])
 
     writer = csv.writer(response)
 
@@ -209,13 +210,9 @@ def analytics_search(request):
     # Note: If we add additional querystring fields, we need to add
     # them to generate_dashboard_url.
     search_happy = request.GET.get('happy', None)
-    search_has_email = request.GET.get('has_email', None)
     search_platform = request.GET.get('platform', None)
     search_locale = request.GET.get('locale', None)
-    search_country = request.GET.get('country', None)
     search_product = request.GET.get('product', None)
-    search_domain = request.GET.get('domain', None)
-    search_api = smart_int(request.GET.get('api', None), fallback=None)
     search_version = request.GET.get('version', None)
     search_query = request.GET.get('q', None)
     search_date_start = smart_date(
@@ -223,27 +220,32 @@ def analytics_search(request):
     search_date_end = smart_date(
         request.GET.get('date_end', None), fallback=None)
     search_bigram = request.GET.get('bigram', None)
+    selected = request.GET.get('selected', None)
+
+    search_has_email = request.GET.get('has_email', None)
+    search_country = request.GET.get('country', None)
+    search_domain = request.GET.get('domain', None)
+    search_api = smart_int(request.GET.get('api', None), fallback=None)
     search_source = request.GET.get('source', None)
     search_campaign = request.GET.get('campaign', None)
     search_organic = request.GET.get('organic', None)
-    selected = request.GET.get('selected', None)
 
     filter_data = []
     current_search = {'page': page}
 
-    search = ResponseMappingType.search()
+    search = ResponseDocType.docs.search()
     f = F()
     # If search happy is '0' or '1', set it to False or True, respectively.
     search_happy = {'0': False, '1': True}.get(search_happy, None)
     if search_happy in [False, True]:
-        f &= F(happy=search_happy)
+        f &= F('term', happy=search_happy)
         current_search['happy'] = int(search_happy)
 
     # If search has_email is '0' or '1', set it to False or True,
     # respectively.
     search_has_email = {'0': False, '1': True}.get(search_has_email, None)
     if search_has_email in [False, True]:
-        f &= F(has_email=search_has_email)
+        f &= F('term', has_email=search_has_email)
         current_search['has_email'] = int(search_has_email)
 
     def unknown_to_empty(text):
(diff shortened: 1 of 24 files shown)


willkg:
This switches us from ElasticUtils to elasticsearch-dsl and also is a complete overhaul of the Elasticsearch using code.

Changes of note:

1. we switched from facets to aggregations (facets are deprecated and will go away soon)
2. we switched from the MoreLikeThis to the morelikethis search query (MoreLikeThis is deprecated and will go away soon)
3. I moved some code around to make the API a little easier to work with
4. I took the time to move the graph that's on the front page dashboard and only shows up if you're an analyzer to the analyzer search.

Things that probably aren't of interest:

1. Reindexing from the command line takes about the same amount of time as it did before. I spent some time tinkering with the code and I'm pretty sure at this point that it's bound on how fast we can get data out of the db. We could speed it up a bit more by getting dicts out instead of model instances, but that requires us to duplicate some code here and there and will make maintenance harder and mistakes more likely.
2. elasticsearch-dsl won't store empty string values (https://github.com/elastic/elasticsearch-dsl-py/issues/202). I ended up writing a `.to_index()` on `ResponseDocType` which both reduces the number of objects created thus speeding indexing up significantly and also gets around the empty string value problem.

To review:

1. read my blog post for some context on the changes: http://bluesock.org/~willkg/blog/mozilla/ditching_elasticutils.html
2. read through the code--half of it is a "this to that" translation and the other half is more involved because I rewrote a bunch of code
3. run the tests
4. run the various `es_*` commands on the command line

fjord/analytics/analyzer_views.py
```diff
@@ -166,12 +167,12 @@ def _analytics_search_export(request, opinions_s):
     keys = Response.get_export_keys(confidential=True)
     total_opinions = opinions_s.count()
 
-    opinions_s = opinions_s.values_list('id')[:MAX_OPINIONS]
+    opinions_s = opinions_s.fields('id')[:MAX_OPINIONS].execute()
```
willkg: The ElasticUtils `values_list` and the elasticsearch-dsl `fields` essentially do the same thing.

With elasticsearch-dsl, you have to explicitly call `execute` to run the search and get back a `Results` instance.

fjord/analytics/analyzer_views.py
```diff
@@ -223,27 +220,32 @@ def analytics_search(request):
     search_date_end = smart_date(
         request.GET.get('date_end', None), fallback=None)
     search_bigram = request.GET.get('bigram', None)
+    selected = request.GET.get('selected', None)
+
+    search_has_email = request.GET.get('has_email', None)
+    search_country = request.GET.get('country', None)
+    search_domain = request.GET.get('domain', None)
+    search_api = smart_int(request.GET.get('api', None), fallback=None)
```
willkg: I moved these down here because this view is almost a copy of the non-analyzer frontpage dashboard. In moving these to this section, it makes it easier to compare the dashboard view with this view to see differences.

fjord/analytics/analyzer_views.py
```diff
@@ -251,57 +253,61 @@ def unknown_to_empty(text):
...
-        f &= F(api=search_api)
+        f &= F('term', api=search_api)
         current_search['api'] = search_api
 
     if search_date_start is None and search_date_end is None:
         selected = '7d'
 
     if search_date_end is None:
-        search_date_end = datetime.now()
+        search_date_end = date.today()
```
willkg: This was a bug. It should never have been a datetime since we're only looking at dates and not time-of-days.

fjord/analytics/analyzer_views.py
```diff
@@ -251,57 +253,61 @@ def unknown_to_empty(text):
...
+    search_date_start, search_date_end = sorted(
+        [search_date_start, search_date_end])
+
     current_search['date_end'] = search_date_end.strftime('%Y-%m-%d')
-    # Add one day, so that the search range includes the entire day.
-    end = search_date_end + timedelta(days=1)
     # Note 'less than', not 'less than or equal', because of the added
     # day above.
-    f &= F(created__lt=end)
+    f &= F('range', created={'lte': search_date_end})
```
willkg: I switched this to an lte after fixing the datetime goof above.

fjord/analytics/analyzer_views.py
```diff
@@ -315,23 +321,25 @@ def unknown_to_empty(text):
...
 
-    search = search.filter(f).order_by('-created')
+    search = search.filter(f).sort('-created')
 
     # If they're asking for a CSV export, then send them to the export
     # screen.
     if output_format == 'csv':
         return _analytics_search_export(request, search)
 
+    original_search = search._clone()
```
willkg: I needed to do two sets of aggregation and thought it was easier to do if I cloned it because one set involves an additional filter or something like that. I need to revisit this.

fjord/analytics/templates/analytics/analyzer/search.html
```diff
@@ -145,6 +145,12 @@
 
 {% block content_middle %}
   <div class="col full">
+    <div class="block graph secret">
+      {# A graph will be dynamically inserted in these divs by Flot. #}
+      <div class="histogram" data-histogram="{{ histogram|to_json }}"></div>
+      <div class="legend"></div>
+      <div class="note">{{ _('Note: All dates and times are in Pacific Time.') }}</div>
+    </div>
```
willkg: The frontpage dashboard used to have a graph on it. A while back, I changed the graph so it only shows up for people who are in the analyzers group.

The use case is that Matt and crew want to be able to do a search and see the graph as a first pass for figuring out whether something is a big deal and needs looking into or not. This helps them field requests from other people about the data.

I left it on the front page because I didn't want to spend the time to move it. In redoing all the Elasticsearch stuff, I decided to finally move it.

fjord/analytics/templates/analytics/dashboard.html
```diff
@@ -128,15 +128,6 @@
...
 {% block content_middle %}
-  {% if (user.is_authenticated() and user.has_perm('analytics.can_view_dashboard')) %}
-    <div class="block graph secret">
-      {# A graph will be dynamically inserted in these divs by Flot. #}
-      <div class="histogram" data-histogram="{{ histogram|to_json }}"></div>
-      <div class="legend"></div>
-      <div class="note">{{ _('Note: All dates and times are in Pacific Time.') }}</div>
-    </div>
-  {% endif %}
-
```
willkg: I moved the graph from the front page dashboard to the analyzer search.

fjord/analytics/tests/test_analyzer_views.py
```diff
@@ -89,6 +89,35 @@ def setUp(self):
...
+        # end date), but what happens is that that includes an extra
+        # day. I suspect there's some funny business in regards to
+        # timezones and we're actually looking at a late time for the
+        # previous day for each day because of timezones and then that
+        # gets handled in flot after being converted to UTC or
+        # something like that.  The point being that "end" is actually
+        # not the end point we want to test against.
+        while d < end:
+            assert d.day in days, 'Day %s has no data.' % d.day
+            d += timedelta(days=1)
```
willkg: Since I moved the graph from the dashboard to the analyzer search, I moved some tests around, too.

fjord/analytics/tests/test_views.py
```diff
@@ -245,25 +245,6 @@ def test_search_format_atom(self):
...
-    #     # Text search
-    #     r = self.client.get(url, {'q': 'relatedlinks', 'format': 'atom'})
-    #     eq_(r.status_code, 200)
-
-    #     assert 'http://www.w3.org/2005/Atom' in r.content
-    #     # FIXME: This is a lousy way to test for a single link with
-    #     # both attributes.
-    #     assert 'rel="related"' in r.content
-    #     assert 'href="http://example.com"' in r.content
-
```
willkg: I deleted this dead code.

fjord/analytics/tests/test_views.py
```diff
@@ -329,36 +310,6 @@ def test_frontpage_index_missing(self):
...
-        # day. I suspect there's some funny business in regards to
-        # timezones and we're actually looking at a late time for the
-        # previous day for each day because of timezones and then that
-        # gets handled in flot after being converted to UTC or
-        # something like that.  The point being that "end" is actually
-        # not the end point we want to test against.
-        while d < end:
-            assert d.day in days, 'Day %s has no data.' % d.day
-            d += timedelta(days=1)
-
```
willkg: This was moved to the analyzer view tests.

fjord/feedback/models.py
```diff
@@ -408,8 +403,103 @@ def infer_platform(cls, product, browser):
         return u''
 
 
-@register_mapping_type
-class ResponseMappingType(FjordMappingType, Indexable):
+class ResponseDocTypeManager(FjordDocTypeManager):
```
willkg: I decided to model this similarly to the Django Model/ModelManager. I'm not sure if that was helpful or just added unneeded complexity.

fjord/search/apps.py
```diff
@@ -0,0 +1,19 @@
...
+
+class SearchConfig(AppConfig):
+    name = 'fjord.search'
+
+    def ready(self):
+        # Set up Elasticsearch connections based on settings
+        urls = getattr(settings, 'ES_URLS', DEFAULT_URLS)
+        timeout = getattr(settings, 'ES_TIMEOUT', DEFAULT_TIMEOUT)
+
+        connections.configure(default={'hosts': urls, 'timeout': timeout})
```
willkg: elasticsearch-dsl maintains connections. So here, we configure all the connections (in our case, it's just "default") so that we can use them elsewhere.

fjord/search/index.py
```diff
@@ -18,7 +25,79 @@
...
+    return doctype
+
+
+def get_es(alias='default'):
+    """Retrieve Elasticsearch instance
+
+    :arg alias: the alias in ES_URLS for this Elasticsearch connection
+
+    """
+    return connections.get_connection(alias=alias)
```
willkg: I kept `get_es` because it's shorter than writing out this line everywhere.

fjord/search/index.py
```diff
@@ -41,34 +120,23 @@ def es_analyze(text, analyzer=None):
...
 
-    return dict((key, val) for key, val in _mapping_types.items()
-                if key in mapping_types)
+    return dict((key, val) for key, val in _doctypes.itesm()
+                if key in doctypes)
 
 
-def get_index():
-    """Returns the index we're using."""
+def get_index_name():
```
willkg: I renamed this because elasticsearch-dsl has an Index class and I decided it was worth clarifying this function.

willkg:
I went through, added a bunch of comments to the PR to make it clearer why things happened and then also fixed some minor things.

fjord/analytics/analyzer_views.py
```diff
@@ -251,57 +253,59 @@ def unknown_to_empty(text):
...
         selected = '7d'
 
     if search_date_end is None:
-        search_date_end = datetime.now()
+        search_date_end = date.today()
     if search_date_start is None:
         search_date_start = search_date_end - timedelta(days=7)
 
+    # If the start and end dates are inverted, switch them into proper
+    # chronoligcal order
```
mythmon: Typo: chronoligcal -> chronological

fjord/analytics/analyzer_views.py
```diff
@@ -251,57 +253,59 @@ def unknown_to_empty(text):
...
+    search_date_start, search_date_end = sorted(
+        [search_date_start, search_date_end])
```
mythmon: Nice! I wouldn't have thought to use sort for that.

fjord/analytics/analyzer_views.py
mythmon: Switching this to dates instead of datetimes makes the code below a lot clearer. :+1:.

fjord/analytics/views.py
```diff
@@ -235,16 +227,16 @@ def unknown_to_empty(text):
...
         current_search['product'] = search_product
 
         if search_version is not None:
             # Note: We only filter on version if we're filtering on
             # product.
-            f &= F(version=unknown_to_empty(search_version))
+            f &= F('term', version=unknown_to_empty(search_version))
             current_search['version'] = search_version
     else:
-        f &= F(product__in=visible_products)
```
mythmon: Is this right? Before you were doing a `__in` query, but now you are doing a normal terms query.

willkg: In Elasticsearch 0.90, the "terms" query was aliased as "in":

https://www.elastic.co/guide/en/elasticsearch/reference/0.90/query-dsl-terms-query.html

I can't tell when they deprecated that alias, but it's not in the new Elasticsearch docs, so I switched it.

willkg:
^^^ Fix typo.

willkg:
@mythmon What do you think here? Is it good enough to land?

mythmon:
I don't really understand the reason to split DocTypes and managers, but I don't really understand the reason behind Django's split either, so that isn't a huge deal. The interface makes sense and looks easy to use on both sides, so :+1:.

Should you add elasticsearch-dsl-py to one of the requirements files? I tried to install it from there, but I didn't find it. I'm also having some trouble with git submodule that I think is something wrong locally. I need to check that out. Either way, I'm not able to actually test this locally right now. I can work on that more this afternoon.

I don't see anything else in the code I want to comment on. The tests and running the code should cover any changes in behavior, so I'm pretty confident in that (assuming I can get it running). The comments and the blog post were very helpful, thanks!

willkg:
I think the split between Model and ModelManager is that ModelManager is an instance of a class that lets you operate on groups of objects whereas Models are about specific things. It lets you do things like have multiple ModelManagers available that think about different groups of objects and in doing that, you get more readable code.

The split here between DocType and DocTypeManager is probably overengineering at this point, but it seemed easier to implement it now this way and see how it feels than it did to just implement DocType and have to potentially fix it later.

I'll look at the requirements file--I probably forgot to do that.

I'm not sure why you're having problems with vendor/, though. Are there error messages?

willkg:
^^^ Rebased against master and also add the requirements.

willkg:
Travis is green. I've addressed all the issues. I did another round of testing. I'm going to call it "good enough" and land it. I'll test extensively on -stage next to make sure I didn't break anything. Then push it to prod either tonight or tomorrow.

willkg:
@mythmon Thank you!

PR #619, opened 2015-07-16
[bug 1177224] Remove dead code

willkg:
This code was just hanging out after we removed the bunch of half-assed
stuff back in commit 7201751.

Quick r?


willkg:
This is all analyzer-area related. None of this is seen by normal users. If the tests pass, it's probably fine.

willkg:
This is all analyzer-facing and the risk is pretty low, so I'm going to self-r+ it.

PR #621, opened 2015-07-27
[bug 1187699] Fix q arg in secret histogram api

willkg:
This adds a test and tweaks some test code to make adding the test easier. The actual fix is to make the `.query()` call correct.

Quick r?


willkg:
This change is benign and only affects a (currently broken) secret histograms API that Robert uses for UA work.

PR #622, opened 2015-07-27
[bug 1187505] Fix feedbackdev persistence code

willkg:
This fixes the feedbackdev persistence code. We upgraded django-waffle recently and they moved some stuff around internally. This code never got updated.

r?


willkg:
This is a trivial fix for something that's pretty much only used by me. I have no idea how someone bumped into it or why they were fiddling with feedbackdev querystring args. This doesn't affect users at all.

PR #624, opened 2015-08-03
[bug 1172019] Merge in new thank you page strings

willkg:
We made changes to the Thank You page strings to make them
product-agnostic. We put that in a separate file that wasn't visible
from the site so as to allow those strings to get translated.

Those strings have been translated in all the priority 1 locales except
Greek (el), but I don't want to keep waiting on this, so I'm pushing it
out now.

r?


willkg:
This PR involves deleting the old file and renaming the new file so that it's got the same filename as the old file. The actual changes were reviewed and approved in pr #598. Travis is green.

Given all that, I'm going to self-r+ this.

PR #626, opened 2015-08-04
[bug 1190356] Redo gengo machine translation

willkg:
Gengo stopped doing synchronous machine translation so I rewrote the
code to do asynchronous translation. Because machine and human
translation are essentially the same now, I unified the two translation
systems and ditched a bunch of code.

r?

### DIFF
--- a/fjord/translations/models.py
+++ b/fjord/translations/models.py
@@ -12,8 +12,6 @@ import waffle
 
 from .gengo_utils import (
     FjordGengo,
-    GengoAPIFailure,
-    GengoMachineTranslationFailure,
     GengoUnknownLanguage,
     GengoUnsupportedLanguage,
 )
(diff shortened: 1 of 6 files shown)


willkg:
There isn't a really good way to *really* test this except in production. Given the human side works in production, I'm going to assume that unifying human/machine here makes it likely the machine side will also work. If not, I'll get error messages and will fix it.

@mythmon Can you review the code changes to make sure they look sane?

fjord/translations/models.py
```diff
@@ -510,9 +434,13 @@ def translate(self, instance, src_lang, src_field, dst_lang, dst_field):
...
 
-        # If the src -> dst isn't a supported pair, log an issue for
-        # metrics purposes and move on.
-        if (lc_src, dst_lang) not in gengo_api.get_language_pairs():
+        # If this is 'standard' tier and if the src -> dst isn't a
+        # supported pair, log an issue for metrics purposes and move
+        # on. We only do this for 'standard' tier since the 'machine'
+        # tier has a different set of pairs which aren't available to
+        # us.
+        if ((self.gengo_tier == 'standard'
```
mythmon: This (and some other things like it) seem like they should be part of the subclasses, instead of mixing all the logic in the super class. Any particular reason the subclasses are so light and the parent class so heavy?

fjord/translations/models.py
```diff
@@ -574,23 +503,29 @@ def push_translations(self):
             # than raise a GengoConfig error.
             return
 
-        balance = gengo_api.get_balance()
-        threshold = settings.GENGO_ACCOUNT_BALANCE_THRESHOLD
+        if self.gengo_watch_balance:
+            balance = gengo_api.get_balance()
+            threshold = settings.GENGO_ACCOUNT_BALANCE_THRESHOLD
```
mythmon: This also seems to only belong to one sub class.

willkg: Testing things with gengo is difficult now and essentially entirely manual. So I was trying to reuse and centralize as much code and do it in a way that minimized code and testing required. Generating machine and human translations is identical except for this balance code which is only needed by the human translation code. I could have the human translation code override this method, do the balance check and then execute the parent method except that I need balance and threshold in the order generation loop.

So I had to choose between this and a more tangled version where I'm passing in termination functions or something like that which I thought was more complex to understand and maintain. I went with simple for now mostly because it eases the burden of testing and maintenance.

fjord/translations/models.py
willkg: I talked about this in my other comment. The gist of it is that testing this is entirely manual and difficult and I'm doing it on a live server so it has monetary consequences. This was the best I could come up with that didn't involve lots of code changes from what I had before, didn't require much extra testing and centralized the code for easier maintenance going forward.

Generally, this code doesn't change over time. We haven't added new features or adjusted existing behavior. The only time this code changes is when Gengo changes some aspect of their API. I'm pretty sure every time they do that, we have to make substantive changes to our code--it's not usually a quick fix.

Having said that, I could do another "flag" thing like I did with the "gengo_watch_balance" flag rather than doing a tier check. I'll look at doing that today.

willkg:
^^^ Rebases against master (with the new django-pipeline changes) and also addresses the "standard" tier check.

mythmon:
The comments about why to have all the logic in the super class make sense. It is nice to only have one kind of translation system now, instead of the sync and async versions.

I can't really test this code, but I stared at it a bit, and it seems sane. I say go for it. r+

willkg:
I'm going to go test it in production. Yay for testing in prod.

Thank you for reviewing!

PR #627, opened 2015-08-05
[bug 1162810] Create AnalyzerProfileFactory to use for creating analy…

lgp171188:
…zers

Fix the existing janes to use AnalyzerProfileFactory so that the code is
clearer in its intent.


willkg:
Awesome! Thank you!

PR #628, opened 2015-08-06
[bug 1191822] Add indexes to heartbeat Answer model

willkg:
This adds indexes to the heartbeat Answer model to make querying faster
for Gregg.

Quick r?


mythmon:
Yep, those are indexes. r+

willkg:
Thank you!

PR #630, opened 2015-08-06
[bug 1191968] Restyle product picker page

willkg:
This restyles the product picker page so it's more like SUMO's. It's
not exactly like SUMO's, but it's enough like SUMO's that we can use the
product icons they use. That'll make some things easier.

Further, this adds product descriptions, but those aren't shown at the
moment because we need to l10n-ize them and we're lacking that
infrastructure.

r?


fjord/feedback/static/css/feedback.less
```diff
@@ -319,51 +319,72 @@ aside {
...
+
+      > li {
+        background: #ccc;
+        border-radius: 6px;
+        box-shadow: 0 2px 4px rgba(0, 0, 0, 0.1);
+        display: inline-block;
+        margin: 0.4rem 0.4rem 0.8rem;
+        position: relative;
+        vertical-align: top;
+        width: 96%;
```
willkg: This is an "eyeballed" number. I couldn't figure out how to get things to line up right.

willkg:
@rehandalal Can you look through this for me?

fjord/feedback/static/css/feedback.less
rehandalal: i would change this to use:
```css
width: calc(100% - 0.8rem);
```

rehandalal:
r+ with that one change

willkg:
^^^ Fixes width. Will wait for Travis and then land.

willkg:
Thank you!

PR #631, opened 2015-08-07
[bug 1174883] Adding Firefox for iOS icon

willkg:
I pulled this icon from SUMO on August 7th. If they change their icon or
it turns out there's a more official official icon or a better looking
non-gif one, then we can switch it out.

Quick r?


willkg:
I'm going to self-r+ this since it's jsut an image that isn't used anywhere, yet. We can always push another one if there are issues.

PR #632, opened 2015-08-10
[bug 1192812] Add support for Firefox for iOS user agents

willkg:
Quick r?

### DIFF
--- a/fjord/base/browsers.py
+++ b/fjord/base/browsers.py
@@ -125,8 +125,8 @@ def parse_ua(ua):
         platform = 'FreeBSD'
     elif platform.startswith('CPU iPhone'):
         # Looks like "CPU iPhone OS 8_3 like Mac OS X" and we want
-        # the "8_3" part.
-        platform_version = platform.split(' ')[3]
+        # the "8_3" part. Then we switch it to "8.3"
+        platform_version = platform.split(' ')[3].replace('_', '.')
         platform = 'iPhone OS'
     elif platform in ('OS X', 'Macintosh'):
         for part in platform_parts:
--- a/fjord/base/browsers.py
+++ b/fjord/base/browsers.py
@@ -66,11 +66,11 @@ def parse_ua(ua):
     no_browser = Browser(UNKNOWN, UNKNOWN, UNKNOWN, UNKNOWN,
                          mobile or None)
 
-    if 'firefox' not in ua.lower():
+    # Firefox browsers have either "Firefox" or "FxiOS" in the UA.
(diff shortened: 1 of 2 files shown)


willkg:
As a side note, this code is insane. I'm tossing around rewriting it as a set of product-specific handler classes and then running a user agent through the classes until one handles it and returns a Browser.

I thought about doing that here, but I need to get this out soon, so I'm spinning that off into a separate bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1192871

willkg:
@mythmon Can you look through this today? The changes are fragile, but I didn't want to do anything more invasive right now.

fjord/base/browsers.py
```diff
@@ -98,11 +98,14 @@ def parse_ua(ua):
...
             browser = 'Iceweasel'
+        elif 'FxiOS' in part:
+            browser = 'Firefox for iOS'
+            browser_version = part[1]
 
     platform = platform_parts.pop(0)
     platform_version = UNKNOWN
 
-    while platform in ['X11', 'Ubuntu', 'U']:
+    while platform in ['X11', 'Ubuntu', 'U', 'iPad', 'iPhone']:
```
mythmon: I imagine that you'll want to know if the user is on an iPhone or an iPad at some point, since FxiOS runs on both and has a fairly distinct UI (afaik). I'm not sure if the data model here supports that though, unfortunately. Maybe keep this in mind when you are doing your big rewrite?

fjord/base/browsers.py
```diff
@@ -120,6 +123,11 @@ def parse_ua(ua):
         platform = 'Linux'
     elif platform.startswith('FreeBSD'):
         platform = 'FreeBSD'
+    elif platform.startswith('CPU iPhone'):
+        # Looks like "CPU iPhone OS 8_3 like Mac OS X" and we want
+        # the "8_3" part.
+        platform_version = platform.split(' ')[3]
```
mythmon: I think you should replace the underscores with periods. iOS versions are generally talked about like `8.3` not `8_3`.

fjord/base/browsers.py
willkg: Yeah... I think I'm not going to worry about that for the rewrite and instead push it off until someone asks for it.

fjord/base/browsers.py
willkg: Mmm... That makes sense. I'll fix that.

willkg:
^^^ Fixes version numbers.

Is this good to go?

mythmon:
lgtm. r+

willkg:
Thank you!

PR #633, opened 2015-08-11
[bug 1193320] Remove eq_ and ok_ from fjord/alerts/

willkg:
r?


willkg:
Travis is green. This is code specific to the test suite. I double checked the new assert statements are correctly structured (i.e. all `eq_` -> `assert` involve a comparison operator and not a comma).

I'm going to self-r+ this. Good work, Will!

PR #634, opened 2015-08-13
[bug 1194189] Fix scaling issue on iOS devices

rehandalal:
pretty simple.

@mythmon  r?

### DIFF
--- a/fjord/feedback/templates/feedback/base.html
+++ b/fjord/feedback/templates/feedback/base.html
@@ -2,7 +2,7 @@
 {% import 'includes/common.html' as common with context %}
 
 {% block extra_headers %}
-  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no" />
 {% endblock %}
 
 {% block site_css %}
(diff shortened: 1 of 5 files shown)


fjord/feedback/templates/feedback/base.html
```diff
@@ -2,7 +2,7 @@
 {% import 'includes/common.html' as common with context %}
 
 {% block extra_headers %}
-  <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no" />
```
willkg: We only need this one. The other ones don't affect feedback forms. So we should ditch those changes from the PR.

willkg: Scratch that. It won't affect the other things, so we should just leave it.

mythmon:
The only device I have to test on is an iPad, which doesn't have the original problem. It doesn't try to zoom to the input field. The page looks the same before and after this patch. With this patch applied I can't zoom in or out though, which I guess is the point.

I'm good with this. r+ from me.

willkg:
Thank you both! :)

PR #635, opened 2015-08-14
[bug 1194450] Remove rating

willkg:
Gradients... we hardly knew you.

Quick r?

### DIFF
--- a/fjord/feedback/migrations/0006_remove_response_rating.py
+++ b/fjord/feedback/migrations/0006_remove_response_rating.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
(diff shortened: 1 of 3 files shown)


willkg:
There was a project to switch from the happy/sad faces to 5 gradations of faces. I started working through the implementation bugs one of which was adding this field to the model. Last quarter, we decided after some thought that we didn't really want to do gradients--it was a lot of work and it wasn't clear what we were getting out of it. Thus we abandoned the project.

This PR removes the column that was added, but never used.

willkg:
@mythmon Can you take a look at this today? It should be straight-forward.

mythmon:
The code changes here look good. Assuming nothing is going to break when those API fields go away (and it sounds like nothing will), I say r+ :boat:.

willkg:
Thank you!

PR #637, opened 2015-08-18
[bug 1195747] Handle "unknown" for country field in hb post

willkg:
This is a stop gap fix to handle "unknown" as a value for the country
field for HB posts. What's going on now is that it throws an HTTP 400
error. This fixes it so that "unknown" gets changed to "UNK" which is a
valid value.

Ultimately we should fix this in HB. There's a PR to fix it there. Once
it's fixed there, we can undo this change.

r?

### DIFF
--- a/fjord/heartbeat/tests/test_api.py
+++ b/fjord/heartbeat/tests/test_api.py
@@ -154,6 +154,32 @@ class HeartbeatPostAPITest(TestCase):
 
         eq_(ans.is_test, False)
 
+    def test_country_unknown(self):
+        """This is a stopgap fix for handling 'unknown' value for country"""
+        survey = SurveyFactory.create()
+
+        data = {
+            'experiment_version': '1',
+            'response_version': 1,
+            'person_id': 'joemamma',
+            'survey_id': survey.name,
+            'flow_id': '20141113',
+            'question_id': '1',
+            'updated_ts': self.timestamp(),
+
+            'question_text': 'ou812?',
+            'variation_id': '1',
+            'country': 'unknown'
+        }
+        resp = self.client.post(
+            reverse('heartbeat-api'),
+            content_type='application/json',
+            data=json.dumps(data))
+
+        eq_(resp.status_code, 201)
+        ans = Answer.objects.latest('id')
+        eq_(ans.country, 'UNK')
+
     def test_survey_doesnt_exist(self):
         """If the survey doesn't exist, kick up an error"""
         data = {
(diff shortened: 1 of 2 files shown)


willkg:
@mythmon This is sort of an emergency fix. Can you look at this soon?

fjord/heartbeat/tests/test_api.py
```diff
@@ -154,6 +154,30 @@ def test_complete_answer(self):
...
+            'person_id': 'joemamma',
+            'survey_id': survey.name,
+            'flow_id': '20141113',
+            'question_id': '1',
+            'updated_ts': self.timestamp(),
+
+            'question_text': 'ou812?',
+            'variation_id': '1',
+            'country': 'unknown'
+        }
```
mythmon: This data looks funny, but I'm sure it matches what it should be. Namely `question_id` being number-in-a-string and `person_id` being an actual string. Not a big deal.

fjord/heartbeat/tests/test_api.py
```diff
@@ -154,6 +154,30 @@ def test_complete_answer(self):
...
+        resp = self.client.post(
+            reverse('heartbeat-api'),
+            content_type='application/json',
+            data=json.dumps(data))
+
+        eq_(resp.status_code, 201)
```
mythmon: I'm sure you know this, but I want to write it down. This is testing only that no errors are thrown, not that the data gets translated to `UNK` or that the data gets kept at all. Which seems fine to me considering the scope of the problem and fix.

mythmon:
I commented on some oddities. They aren't bad, I just want to make sure I am interpreting things correctly. Assuming my understanding is correct, r+.

fjord/heartbeat/tests/test_api.py
willkg: The API is documented here if that helps:

http://fjord.readthedocs.org/en/latest/hb_api.html#required

fjord/heartbeat/tests/test_api.py
willkg: That's a good point. I'll add checks for those things.

fjord/heartbeat/tests/test_api.py
mythmon: Can confirm. This matches the docs.

Do you do this a lot? In a different, not-an-emergency time, maybe it would make sense to use FactoryBoy to create api data? As I understand it, it can also generate arbitrary objects too.

willkg:
^^^ That fixes the test. I'll wait for Travis and then land it.

willkg:
Thank you!

fjord/heartbeat/tests/test_api.py
willkg: Yeah, that's a good idea. FactoryBoy or a `generate_minmal` function or something equivalent.

PR #638, opened 2015-08-18
[bug 1195899] Upgrade django to 1.7.10

willkg:
Quick r?


willkg:
@mythmon Can you do a quick r? on this? Then I can do SUMO.

mythmon:
Confirmed that the git commit is correct on django/django, and that the hash matches what I downloaded from PyPI.

r+

PR #640, opened 2015-08-19
Move SUMOSuggest into separate module

willkg:
This is somewhat cosmetic. This structure makes it easier to have more
complicated suggestion providers.

r?


willkg:
@mythmon Can you look at this?

I know we discussed filesystem/python-module structure way back when I did the original suggestions stuff. The trigger suggester is way more complicated and involves db models, so it really needs to be in its own module. I figured I'd move this one to see how it feels and whether there are other things that should be done, too.

Thoughts?

fjord/suggest/providers/sumo/tests/test_sumosuggest.py
```diff
@@ -13,18 +13,18 @@
...
     SUMO_SUGGEST_API_URL,
     SUMO_SUGGEST_SESSION_KEY,
-    SUMOSuggest
 )
 from fjord.suggest.tests import SuggesterTestMixin
 from fjord.redirector.tests import RedirectorTestMixin
 
 
-class SUMOSuggestTestCase(TestCase):
+class SumoTestCase(TestCase):
```
mythmon: Nit: You use the all caps SUMO everywhere else, but the title case Sumo here (personally I prefer title case).

willkg: My bad. I did a lousy search-and-replace and then instead of undoing it, I went and "fixed things" by hand.

SUMO is an acronym, so it's annoying but I figured SUMO is clearer than Sumo.

mythmon:
This seems fine to me, and will probably help a lot with the more complex services. I don't actually remember much about the last time we talked about this, but nothing jumps out at me here.

willkg:
^^^ Fixes test names. Once Travis is good, I'll land it.

willkg:
Thank you!

PR #641, opened 2015-08-20
[bug 1192812] Fix firefox for ios inferring

willkg:
What's going on is that I had handling for iPhone and iPad, but not iPod touch. I fixed that and also added some tests to verify that inferring works correctly.

r?


willkg:
The actual change here is a one-line change that's pretty benign in that it either affects things for good or it doesn't do anything at all. The rest of this PR is additional tests for iPod touch.

Given that and that Travis is green, I'm going to self-r+ this.

PR #642, opened 2015-08-20
[bug 1192812] Fix platform inferring for firefox for ios

willkg:
I missed this in the previous fix. Bah.

I'll wait for Travis to greenlight it and then self-r+ it and land it.


willkg:
It's a minor change with tests. Travis is green. I'm going to self-r+ this.

PR #643, opened 2015-08-20
[bug 1175123] Initial implementation of Trigger suggester

willkg:
Implements the Trigger suggester which allows analyzers to set up
trigger rules for adding suggestions to feedback that match those rules.

r?

### DIFF
--- a/fjord/base/forms.py
+++ b/fjord/base/forms.py
@@ -1,6 +1,40 @@
 from django.forms import fields
+from django.forms import widgets
 
 from fjord.base.validators import EnhancedURLValidator
+
+
+class StringListField(fields.CharField):
+    widget = widgets.Textarea()
+
+    def prepare_value(self, value):
+        """Converts a value into a value appropriate for a Textarea widget
+
+        :arg value: None or a list of unicode strings
+
+        :returns: unicode string
+
+        """
+        if not value:
+            value = u''
+        else:
+            value = u'\n'.join(value)
+        return value
+
+    def clean(self, value):
+        """Converts a value from the HTML form to a ListField value
+
+        This also strips each individual unicode to remove space
+        before and after the value.
+
+        :arg value: unicode string
+
+        :returns: list of unicodes
+
+        """
+        values = super(StringListField, self).clean(value)
+        val = [val.strip() for val in values.splitlines() if val]
+        return val
 
 
 class EnhancedURLField(fields.URLField):
--- a/fjord/base/models.py
+++ b/fjord/base/models.py
@@ -1,6 +1,8 @@
+import ast
 import json
 
 from django.contrib.auth.models import User
+from django.core.exceptions import ValidationError
 from django.db import models
 from django.utils.translation import ugettext_lazy as _lazy
 
@@ -14,6 +16,56 @@ ModelBase = models.Model
 
 class Profile(models.Model):
     user = models.OneToOneField(User, on_delete=models.CASCADE)
+
+
+class ListField(models.TextField):
+    description = _lazy('List of values')
+
+    __metaclass__ = models.SubfieldBase
+
+    def __init__(self, *args, **kwargs):
+        super(ListField, self).__init__(*args, **kwargs)
+
+    def to_python(self, value):
+        """Converts the value to a Python value
+
+        :arg value: None, list or unicode from the db
+
+        :returns: list of things
+
+        """
+        if not value:
+            value = []
+
+        if isinstance(value, list):
+            return value
+
+        try:
+            # This takes a unicode and reconsistutes it into Python base
+            # types.
+            value = ast.literal_eval(value)
+            assert isinstance(value, list)
+            return value
+        except Exception as exc:
+            # Need to return a ValidationError, but this throws a
+            # SyntaxError, so we (poorly) wrap it.
+            raise ValidationError(repr(exc))
+
+    def get_prep_value(self, value):
+        """Perform preliminary non-db specific value checks and conversions"""
+        if value is None:
+            return value
+
+        return unicode(value)
+
+    def value_to_string(self, obj):
+        """Returns a string value of this field from the passed obj
+
+        Used by the serialization framework.
+
+        """
+        value = self._get_val_from_obj(obj)
+        return self.get_db_prep_value(value)
 
 
 class EnhancedURLField(models.CharField):
--- a/fjord/base/tests/test_models.py
+++ b/fjord/base/tests/test_models.py
@@ -1,0 +1,38 @@
+from django.core.exceptions import ValidationError
+
+from fjord.base.models import ListField
+from fjord.base.tests import TestCase
+
+
+class ListFieldTestCase(TestCase):
+    def test_to_python(self):
+        tests = [
+            # test data, expected
+            (None, []),
+            ([], []),
+            ([1, 2, 3], [1, 2, 3]),
+            (u'[1, 2, 3]', [1, 2, 3]),
+            (u"[u'a', u'b', u'c']", [u'a', u'b', u'c'])
+        ]
+        field = ListField()
+        for testcase, expected in tests:
+            assert field.to_python(testcase) == expected
+
+    def test_to_python_non_list_raises_validationerror_on_assert(self):
+        field = ListField()
+        self.assertRaises(ValidationError, field.to_python, (u'42',))
+
+    def test_to_python_raises_validationerror_on_syntaxerror(self):
+        field = ListField()
+        self.assertRaises(ValidationError, field.to_python, (u'abc',))
+
+    def test_get_prep_value(self):
+        tests = [
+            # test data, expected
+            ([], u'[]'),
+            ([1, 2], u'[1, 2]'),
+            ([u'a', u'b'], u"[u'a', u'b']")
+        ]
+        field = ListField()
+        for testcase, expected in tests:
+            assert field.get_prep_value(testcase), expected
--- a/fjord/redirector/__init__.py
+++ b/fjord/redirector/__init__.py
@@ -1,4 +1,5 @@
-__all__ = ['build_redirect_url', 'get_redirectors', 'Redirector']
+__all__ = ['build_redirect_url', 'get_redirectors', 'Redirector',
+           'RedirectParseError']
 
 
 default_app_config = 'fjord.redirector.apps.RedirectorConfig'
@@ -17,4 +18,4 @@ def get_redirectors():
     return list(_REDIRECTORS)
 
 
-from .base import build_redirect_url, Redirector
+from .base import build_redirect_url, Redirector, RedirectParseError
--- a/fjord/suggest/providers/trigger/models.py
+++ b/fjord/suggest/providers/trigger/models.py
@@ -1,0 +1,197 @@
+import re
+
+from django.db import models
+from django.utils.lru_cache import lru_cache
+
+from fjord.base.models import ListField, ModelBase
+from fjord.feedback.models import Product
+
+
+class TriggerRuleManager(models.Manager):
+    def all_enabled(self):
+        return self.filter(is_enabled=True).order_by('sortorder')
+
+
+@lru_cache()
+def _generate_keywords_regex(keywords):
+    """Generates a Regex object to compare with
+
+    * for each keyword
+      * converts white space to '\s+'
+    * joins the parts to create a middle
+    * joins that with empty string bounds
+    * compiles it into a Regex and returns it
+
+    :arg keywords: list of unicode strings
+
+    :returns: regex object
+
+    """
+    # Make absolutely sure it's a list of unicode strings because
+    # crossing the streams is bad.
+    assert [isinstance(keyword, unicode) for keyword in keywords]
+
+    def _escape(utext):
+        """Escape regexp characters
+
+        This is different than re.escape which escapes everything that
+        isn't an alphanumeric character which doesn't work well with
+        unicode and also escapes spaces which I don't want because
+        we want to convert consecutive spaces to \s+ later.
+
+        """
+        chars = r'\/.^$*+?{}()|?<>[]-'
+        for c in chars:
+            utext = utext.replace(c, u'\\' + c)
+        return u''.join(utext)
+
+    keywords = [
+        re.sub(ur'\s+', ur'\s+', _escape(keyword))
+        for keyword in keywords
+    ]
+    middle = (
+        u'(?:' +
+        u'|'.join(keywords) +
+        u')'
+    )
+    regex = ur'((?:^|\s)' + middle + ur'(?:\s|$))'
+    return re.compile(regex, re.LOCALE | re.IGNORECASE | re.UNICODE)
+
+
+class TriggerRule(ModelBase):
+    # Trigger properties
+    slug = models.SlugField(
+        unique=True,
+        help_text=(
+            u'Required: One-word unique slug for this trigger rule. Used '
+            u'in redirection urls and GA event tracking. Users will see '
+            u'this. Do not change once it is live!'
+        )
+    )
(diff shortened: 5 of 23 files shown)


willkg:
This is described at: https://wiki.mozilla.org/Firefox/Input/Trigger_suggester

There are a few things that aren't done yet that will be done under different bugs. I figured it's worth splitting this out into parts and this is the minimum viable initial part.

Outstanding things that will be covered by other bugs:

1. templated urls
2. trigger rule tester
3. list of available locales for trigger rule form

To review:

1. review the code for oddities
2. run the tests

I've gone through and done the hand testing of creating rules and then creating feedback that matches and doesn't match rules to make sure that works. If you want to do that, too, that's cool, but don't feel like you have to. We'll push this out and if there are issues we didn't catch in review, we can fix them as we find them.

willkg:
@mythmon Can you work through this? It's like 900 lines of code, but some of that is boilerplate and similar to the SUMO suggester. Sometime in the next week would be cool.

fjord/redirector/__init__.py
```diff
@@ -17,4 +17,4 @@ def get_redirectors():
     return list(_REDIRECTORS)
 
 
-from .base import build_redirect_url, Redirector
+from .base import build_redirect_url, Redirector, RedirectParseError
```
mythmon: Is this a "import for export" type of thing? I don't see it in the `__all__` definition above.

willkg: Oops. It should be in the `__all__`.

fjord/suggest/providers/trigger/models.py
```diff
@@ -0,0 +1,253 @@
...
+      * converts white space to '\s+'
+    * joins the parts to create a middle
+    * joins that with empty string bounds
+    * compiles it into a Regex and returns it
+
+    :arg keywords: list of words and phrases to match
+
+    :returns: regex object
+    """
+    assert isinstance(keywords, unicode)
```
mythmon: Here you assert `keywords` is a unicode, but in the docstring you call it a list.

fjord/suggest/providers/trigger/models.py
```diff
@@ -0,0 +1,253 @@
...
+
+    def _escape(utext):
```
mythmon: The point of this method is to make strings safe for insertion into regex, right? Would `re.escape` work as well?

fjord/suggest/providers/trigger/models.py
```diff
@@ -0,0 +1,253 @@
...
+        for part in keywords.splitlines()
+    ]
+    if len(parts) == 1:
+        middle = parts[0]
+    else:
+        middle = (
+            u'(?:' +
+            u'|'.join(parts) +
+            u')'
+        )
```
mythmon: I think it would be better to always do the group+join part. It would be safe to do in the case of a one-length-list, and it is more consistent (also less code).

fjord/suggest/providers/trigger/models.py
```diff
@@ -0,0 +1,253 @@
...
+    regex = u'((?:^|\\s)' + middle + u'(?:\\s|$))'
```
mythmon: Here (and throughout this function) you could use raw unicode strings like `ur'((?:^|\s)'`.

fjord/suggest/providers/trigger/models.py
```diff
@@ -0,0 +1,253 @@
...
+        help_text=(
+            u'Required: One-word unique slug for this trigger rule. Used '
+            u'in redirection urls and GA event tracking. Users will see '
+            u'this. Do not change once it is live!'
+        )
+    )
+    title = models.CharField(
+        max_length=255,
+        help_text=(
+            u'RequireD: Title of the suggestion--best to keep it a short '
```
mythmon: Typo: RequireD

fjord/suggest/providers/trigger/models.py
```diff
@@ -0,0 +1,253 @@
...
+    url = models.URLField(
+        help_text=(
+            u'Required: URL for the suggestion. This should be '
+            u'locale-independent if possible.'
+        )
+    )
+    sortorder = models.IntegerField(
+        default=5,
+        help_text=(
+            u'RequireD: Allows you to dictate which trigger rules are more '
```
mythmon: type: RequireD here too

fjord/suggest/providers/trigger/models.py
willkg: Actually, `re.escape` doesn't do what I want. It escapes everything that's not an ascii alphanumeric character which makes a mess of things because I want to do different things with spaces and it doesn't work well with utf-8.

This probably isn't perfect, but it covers all the characters I suspect someone writing a trigger rule would need covered.

fjord/suggest/providers/trigger/models.py
willkg: Oh, neat. It never occurred to me there was a ur thing.

fjord/suggest/providers/trigger/models.py
willkg: That makes sense. I'll do that.

fjord/suggest/providers/trigger/models.py
```diff
@@ -0,0 +1,253 @@
...
+        # Check if this rule matches all locales
+        if not self.locales:
+            return True
+
+        # Check if there was a locale supplied
+        if not locale:
+            return False
+
+        # Check if the locale was in the rule
+        return locale in self.locales.split(',')
```
mythmon: If you took the behavior of the `locales` field, and made it into its own special kind of field that stored multiple values as a comma-delimited string, converting to a string when it goes into the database, and converting into a list when it comes out, I'd be quite happy, and I would steal it to use on Edwin. I'm doing a similar thing there with ad-hoc methods like this.

That would be hard though, and the ad-hoc way works fine.

fjord/suggest/providers/trigger/models.py
```diff
@@ -0,0 +1,253 @@
...
+        return product_name in product_db_names
+
+    def match_description(self, description):
+        if not self.keywords:
+            return True
+
+        if not description:
+            return False
+
+        assert isinstance(description, unicode)
```
mythmon: +1 to putting assertions in the code, btw.

fjord/suggest/providers/trigger/provider.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+def format_redirect(slug):
+    return '{0}.{1}'.format(TRIGGER, slug)
+
+
+def parse_redirect(redirect):
+    try:
+        trigger, slug = redirect.split('.')
+        return trigger, slug
```
mythmon: Why not simply `return redirect.split('.')`. Is that just for clarity?

mythmon: Actually, I take it back. Having the binding first makes it assert that there are exactly 2 parts, which is useful.

fjord/suggest/providers/trigger/models.py
willkg: Plus it's the worst kind of typo that requires me to rebuild the initial migration. Boo hiss.

fjord/suggest/providers/trigger/models.py
willkg: It's funny you mention this. I'm almost positive I've written a CommaSeparatedStringField which sort of does the same thing as CommaSeparatedIntegerField, but it works with strings (plus CommaSeparatedIntegerField doesn't work the way I expected--it's weird). I can't remember which project I wrote it for. It's also possible I wrote it and either didn't use it or deleted that code at some point.

I'll toy with writing one and see whether having it is a good idea or not.

Even with it, I have code for collapsing pre/post spaces. So I'm not wholly sure whether it'd really clean things up here.

willkg: Maybe something like this: http://brunorocha.org/python/django/django-listfield-e-separetedvaluesfield.html

fjord/suggest/providers/trigger/models.py
willkg: This particular issue kept kicking my ass.

fjord/suggest/providers/trigger/provider.py
willkg: This is because the code I copied it from converted the second part into an int and had exception handling for that. I didn't think of the fact that this enforces it to be two parts--that's a good reason to leave it as is.

fjord/suggest/providers/trigger/provider.py
```diff
@@ -0,0 +1,112 @@
...
+class TriggerRedirector(Redirector):
+    """Provides redirection urls"""
+    def handle_redirect(self, request, redirect):
+        # If it's not a trigger redirect, return.
+        if not redirect.startswith(TRIGGER):
+            return
+
+        # Get the response id for the last feedback left.
+        response_id = request.session.get('response_id', None)
+        if response_id is None:
```
mythmon: Do all the redirectors do this? Is it too late to make the response a parameter that this function takes, instead of having to pull it out of the session?

fjord/suggest/providers/trigger/templates/trigger/triggerrule_list.html
```diff
@@ -0,0 +1,46 @@
...
+          <td>{{ obj.title }}</td>
+          <td>{{ obj.locales or '--' }}</td>
+          <td>
+            {% for prod in obj.products.all() %}
+              {{ prod.db_name }}{% if not loop.last %}, {% endif %}
+            {% else %}
+              --
+            {% endfor %}
+          </td>
+          <td>{{ obj.versions or '--' }}</td>
```
mythmon: In the body, the order is locales, products, versions. In the headers it is versions, locales, products. I think versions got misplaced in one of those.

fjord/suggest/providers/trigger/provider.py
willkg: The redirector system isn't feedback specific. I wrote it to be general to any parts of Input. I figured that makes it easier to extract and reuse elsewhere.

However, the two places it's currently being used are both redirecting links from the thank you page which is post-feedback and thus there will always be a response id in the session. I could extract that code out and make it a utility method in the feedback module somewhere. I'm not sure whether that really helps a lot here, but you're right--both of the redirectors have the same boilerplate code right now.

fjord/suggest/providers/trigger/tests/test_models.py
```diff
@@ -0,0 +1,189 @@
...
+            ('38', '38.0.5', False),
+
+            ('38*', '', False),
+            ('38*', '38', True),
+            ('38*', '38.0.5', True),
+
+            ('38*,37.1', '', False),
+            ('38*,37.1', '38', True),
+            ('38*,37.1', '38.0.5', True),
+            ('38*,37.1', '37', False)
```
mythmon: I think you should add a case here `('38*,37.1', '37.1', True)`, to make sure it can match with not-the-first version.

fjord/suggest/providers/trigger/tests/test_models.py
```diff
@@ -0,0 +1,189 @@
...
+    def test_keyword_regexes_are_cached(self):
+        for text in [u'cat', u'dog', u'mouse']:
+            tr = TriggerRuleFactory(keywords=text)
+            tr.match_description(text)
+
+            tr = TriggerRuleFactory(keywords=text)
+            tr.match_description(text)
+
+        cache_info = _generate_keywords_regex.cache_info()
+        assert cache_info.hits >= 3
```
mythmon: I've never seen a test like this, though I can imagine why it is important that these regexes are cached. Did you run into this in development? Or just wanted to be very sure this would work well?

fjord/suggest/providers/trigger/templates/trigger/triggerrule_list.html
willkg: Bah.

fjord/suggest/providers/trigger/tests/test_models.py
willkg: This test is mostly to make sure the code is correct and that the regex objects are getting cached.

The suggestion providers need to be *fast*. Caching in memory (as opposed to memcache) will help with reducing the work a bit.

mythmon:
That's all I've got for today. I'll do another quick check tomorrow, and pull it down and run it.

This looks good. It seems like the suggestion/redirect provider stuff you did paid off pretty well.

willkg:
^^^ Fixes based on your comments.

willkg:
Er, that has fixes based on your comments except for writing up a comma-separated string field. I'll fiddle with that today to see if it's a good value.

willkg:
^^^ Rewrote the code using ListField (model field) and StringListField (form field). I like it. I think it's easier to work with than the previous iteration.

@mythmon Can you look at that?

fjord/base/forms.py
```diff
@@ -1,8 +1,26 @@
 from django.forms import fields
+from django.forms import widgets
 
 from fjord.base.validators import EnhancedURLValidator
 
 
+class StringListField(fields.CharField):
+    widget = widgets.Textarea()
+
+    def prepare_value(self, value):
```
mythmon: Could you add docstrings to these methods to note which of these methods deal with from-the-database data and which ones deal with from-python-objects data?

fjord/base/models.py
```diff
@@ -16,6 +18,47 @@ class Profile(models.Model):
...
+            value = []
+
+        if isinstance(value, list):
+            return value
+
+        try:
+            # This takes a unicode and reconsistutes it into Python base
+            # types.
+            return ast.literal_eval(value)
+        except Exception as exc:
```
mythmon: Can you catch SyntaxError here, instead of all Exceptions?

fjord/base/models.py
```diff
@@ -16,6 +18,47 @@ class Profile(models.Model):
...
+        if not value:
+            value = []
+
+        if isinstance(value, list):
+            return value
+
+        try:
+            # This takes a unicode and reconsistutes it into Python base
+            # types.
+            return ast.literal_eval(value)
```
mythmon: Should you check that this is a list after it has been evaled?

fjord/base/tests/test_models.py
```diff
@@ -0,0 +1,34 @@
...
+from fjord.base.tests import TestCase
+
+
+class ListFieldTestCase(TestCase):
+    def test_to_python(self):
+        tests = [
+            # test data, expected
+            (None, []),
+            ([], []),
+            ([1, 2, 3], [1,2 ,3]),
```
mythmon: Typo in the expected value. bad spacing around the 2.

fjord/base/models.py
willkg: I don't know if SyntaxError is the only possible exception that can be thrown here. Regardless, I think I want to treat all exceptions as ValidationErrors, so seems better to keep it like this.

fjord/base/models.py
willkg: Yeah, that's a good idea.

willkg:
^^^ Addresses your last set of comments:

* add assertion and test for list-ish-ness
* add docstrings (some of which copied from the django docs because might as well)
* remove a method that wasn't used (no clue why i added it in the first place (which is not confidence-creating))
* remove errant print statements

mythmon:
This looks good. I got the tests working before, and they all passed. The new model and form fields stuff looks well enough tested, and only a little goofy. I didn't test it myself, because I couldn't get Persona to work locally (I should fix that), but I'm sure we can test it in prod.

r+

willkg:
^^^ That squashed all the commits into one. Travis is green. I'll go through it once more, but I think we're good to go now.

willkg:
^^^ Removes a function that was never used. I'll wait for Travis then land it.

willkg:
LANDED!

Thank you so much for reviewing this!

PR #652, opened 2015-08-25
[bug 1193330] Remove eq_ and ok_: search

staranjeet:
files under search directory. It replaces eq_ and ok_ with proper
assert statements.


willkg:
Awesome!

PR #653, opened 2015-08-26
[bug 1175123] Fix keyword matching

willkg:
Keywords wouldn't match if bounded by a non-alphnum character. e.g.
"best" would not match "firefox is the best!" because the "!" was
neither an end-of-string nor a whitespace character.

This changes the bounds to \W and adds a test.

r?

### DIFF
--- a/fjord/suggest/providers/trigger/models.py
+++ b/fjord/suggest/providers/trigger/models.py
@@ -54,8 +54,8 @@ def _generate_keywords_regex(keywords):
         u'|'.join(keywords) +
         u')'
     )
-    regex = ur'((?:^|\s)' + middle + ur'(?:\s|$))'
-    return re.compile(regex, re.LOCALE | re.IGNORECASE | re.UNICODE)
+    regex = ur'((?:^|\W)' + middle + ur'(?:\W|$))'
+    return re.compile(regex, re.IGNORECASE | re.UNICODE)
 
 
 class TriggerRule(ModelBase):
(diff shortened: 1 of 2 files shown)


willkg:
@mythmon Can you look at this? It's very small.

mythmon:
Would the regex character `\b` (word boundary) work? I think it would simplify this regex to just `ur'(\b' + middle + ur'\b)'`

willkg:
I spent a day trying to get `\b` to work with unicode and I'm not really sure why it didn't.

http://www.standu.ps/status/26411

I'll try it again now. It's possible it was related to some other issue that has since been fixed.

mythmon:
Oh, ok. Does the flag `re.UNICODE` help?

mythmon:
Also, for what it is worth, this change as it is looks fine. The test accurately tests what you are trying to fix, and the change seems to have fixed it. I don't know that `\b` works well with unicode or is even appropriate.

That being said, maybe it would be good to add a unicode test case, if you are having trouble with that? I tried a couple of cases, and it seems to work for me.

willkg:
It's got the `re.UNICODE` flag already, so I'm not sure what you're suggesting I should change.

willkg:
There's a unicode test case already:

https://github.com/mozilla/fjord/blob/master/fjord/suggest/providers/trigger/tests/test_models.py#L84

Plus I've been doing my manual testing with unicode characters.

willkg:
^^^ That tweaks the regex (ditches the match group since we're using `.search()`) and adds a note about `\b`.

mythmon:
I think this is probably fine. Easy to change if something doesn't work quite right. r+

willkg:
Thank you!

PR #654, opened 2015-08-28
[bug 1193331] Remove eq_ and ok_: suggest.

staranjeet:
files under suggest/providers directory. It replaces eq_ and ok_ with proper
assert statements. This directory was left out while submitting PR #651 .


willkg:
Looks great! Thank you!

PR #655, opened 2015-08-28
1193325 remove eq ok

staranjeet:
files under feedback directory. It replaces eq_ and ok_ with proper 
assert statements.


willkg:
There's something fishy about this pull request. It's got two commits in it one of which looks like a merge commit.

Can you create a new branch off of the current master tip, then cherry-pick 291e6a828b21e0540ce4a99eca67b067d8eeb2f6 which I think is the commit that makes the changes you're trying to make?

staranjeet:
submitted #656

PR #658, opened 2015-08-29
1199324 tests

willkg:
Fixes the product picker test.

Then while I was futzing with the smoke tests, I also removed unittestzero because it's superfluous and then I updated the requirements.


willkg:
This code isn't tested by Travis and doesn't affect our server environments at all.

I'll review it on Monday and if it looks good, I'll merge it.

willkg:
^^^ Tweaked a couple of things. It's good to go now.

PR #659, opened 2015-09-01
[bug 1032951] Upgrade lxml to 3.4.4

willkg:
This only affects local development environments and doesn't affect server environments.

I have a bug open with webops to figure out the server environments side.

The only two things that seem to use lxml are pyquery (which we use in the tests) and translate-toolkit (which we use extracting strings). I don't see evidence of other things using lxml. It's entirely possible that nothing of substance uses it at all.

Given that, I figured I might as well make a PR for it.

r?


willkg:
I verified the sha. I'm just going to merge this in.

PR #660, opened 2015-09-02
[Bug 1198766] tester

willkg:
This lets trigger rule writers test the trigger rule against recent feedback.

r?

### DIFF
--- a/fjord/suggest/providers/trigger/api_views.py
+++ b/fjord/suggest/providers/trigger/api_views.py
@@ -1,0 +1,106 @@
+from django.utils.decorators import method_decorator
+
+import rest_framework.response
+import rest_framework.views
+from rest_framework import exceptions
+from rest_framework import serializers
+from rest_framework.authentication import SessionAuthentication
+
+from fjord.base.api_utils import StrictArgumentsMixin
+from fjord.base.utils import (
+    analyzer_required,
+    cors_enabled,
+)
+from fjord.feedback.models import Response as FeedbackResponse
+from fjord.suggest.providers.trigger.models import TriggerRuleMatcher
+
+
+class TriggerRuleTestSerializer(StrictArgumentsMixin, serializers.Serializer):
+    locales = serializers.ListField(
+        child=serializers.CharField(trim_whitespace=True, allow_blank=False))
+    products = serializers.ListField(
+        child=serializers.CharField(trim_whitespace=True, allow_blank=False))
+    versions = serializers.ListField(
+        child=serializers.CharField(trim_whitespace=True, allow_blank=False))
+    keywords = serializers.ListField(
+        child=serializers.CharField(trim_whitespace=True, allow_blank=False))
+    url_exists = serializers.NullBooleanField()
+
+
+class CSRFLessSessionAuthentication(SessionAuthentication):
+    def enforce_csrf(self, request):
+        return
+
+
+class TriggerRuleMatchViewAPI(rest_framework.views.APIView):
+    authentication_classes = (CSRFLessSessionAuthentication,)
+
+    @classmethod
+    def as_view(cls, **initkwargs):
+        # Enable CORS
+        view = super(TriggerRuleMatchViewAPI, cls).as_view(**initkwargs)
+        return cors_enabled('*', methods=['POST'])(view)
+
+    def rest_error(self, errors):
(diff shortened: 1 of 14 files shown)


willkg:
To review:

1. read through the code
2. run the tests (or let travis run them--either is fine)
3. suggest fixes for the mediocre js code

@mythmon Can you look through this sometime this week?

mythmon:
Travis doesn't like this :bomb:.

willkg:
It was a network failure on setup. I kicked it off again.

willkg:
^^^ Everything passed.

fjord/suggest/providers/trigger/api_views.py
```diff
@@ -0,0 +1,92 @@
...
+
+class TriggerRuleMatchViewAPI(rest_framework.views.APIView):
+    authentication_classes = (CSRFLessSessionAuthentication,)
+
+    @classmethod
+    def as_view(cls, **initkwargs):
+        # Enable CORS
+        view = super(TriggerRuleMatchViewAPI, cls).as_view(**initkwargs)
+        view = cors_enabled('*', methods=['POST'])(view)
+        return view
```
mythmon: Why the assign-then-return here?

mythmon:
Oh, I forgot that adding comments to commits instead of the diff here makes things annoying. Sorry about that.

fjord/suggest/providers/trigger/api_views.py
willkg: Mmmm... I had this return the value at 28, then later spent time fiddling with `csrf_exempt` and another decorator and switched it to this form which was easier to read. I never switched it back.

I'll fix it.

willkg:
I'm totally fine with adding comments to the individual commits. The notifications go to a different email address (yay github!), but it's fine by me.

willkg:
Yes. It's a link to seeing a single day on the front page dashboard.

willkg:
^^^ That redoes the api endpoint to use serializers, addresses your code comments so far and cleans up the JS.

willkg:
^^^ And that switches to `Array.map`. I think that covers everything.

willkg:
^^^ Nixes some dead code. Less to review now!

fjord/suggest/providers/trigger/api_views.py
```diff
@@ -0,0 +1,106 @@
...
+from fjord.base.api_utils import StrictArgumentsMixin
+from fjord.base.utils import (
+    analyzer_required,
+    cors_enabled,
+)
+from fjord.feedback.models import Response as FeedbackResponse
+from fjord.suggest.providers.trigger.models import TriggerRuleMatcher
+
+
+class TriggerRuleTestSerializer(StrictArgumentsMixin, serializers.Serializer):
```
mythmon: Ooh, I forgot about `StrictArugmentsMixin`. So cool

fjord/suggest/providers/trigger/tests/test_api.py
```diff
@@ -0,0 +1,336 @@
...
+            'url_exists': None
+        }
+        resp = self.client.post(
+            reverse('triggerrule-match'),
+            content_type='application/json',
+            data=json.dumps(data)
+        )
+        assert resp.status_code == 200
+        # This matches everything because it's an empty rule. We'll
+        # verify the count and assume it's correct.
```
mythmon: This comment doesn't match the code anymore.

mythmon:
Ok. I think this is good now. Well, not good, but as good as it is going to be. I don't see any glaring issues, and it looks better than when it started. It might be good to handle the error from the ajax call (where you already have a TODO), but I'm not sure what you would do. Maybe `alert("It broke. Sorry.")`.

r+

fjord/suggest/providers/trigger/tests/test_api.py
willkg: Ahh... you're right. Oops.

willkg:
^^^ Fixes the comment and the adds a mediocre error "handler".

I'll wait for Travis to ok this, then land it.

Thank you so much!

willkg:
Yay Travis!

willkg:
Side note: Fixing the bundles and static files is covered in another bug. Right now it's a big mess. :(

PR #661, opened 2015-09-04
[bug 1198757] Add templated urls to trigger rule suggester

willkg:
This also adds tests for the redirector which I apparently didn't write the first time around. :(

r?

### DIFF
--- a/fjord/suggest/providers/trigger/provider.py
+++ b/fjord/suggest/providers/trigger/provider.py
@@ -1,6 +1,8 @@
 import logging
+import urllib
 
 from fjord.base.google_utils import ga_track_event
+from fjord.feedback.models import Response as FeedbackResponse
 from fjord.redirector import (
     Redirector,
     RedirectParseError,
@@ -37,6 +39,20 @@ def build_ga_category(rule):
     return 'trigger_{rule.slug}_{rule.id}'.format(rule=rule)
 
 
+def interpolate_url(url, response):
+    if '{' not in url:
+        return url
+
+    data = {
+        'LOCALE': urllib.quote_plus(response.locale.encode('utf-8')),
+        'PRODUCT': urllib.quote_plus(response.product.encode('utf-8')),
+        'VERSION': urllib.quote_plus(response.version.encode('utf-8')),
+        'PLATFORM': urllib.quote_plus(response.platform.encode('utf-8')),
+        'HAPPY': 'happy' if response.happy else 'sad',
+    }
+    return url.format(**data)
+
+
 class TriggerRedirector(Redirector):
     """Provides redirection urls"""
     def handle_redirect(self, request, redirect):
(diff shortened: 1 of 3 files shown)


fjord/suggest/providers/trigger/provider.py
```diff
@@ -37,6 +39,20 @@ def build_ga_category(rule):
...
+        return url
+
+    data = {
+        'LOCALE': urllib.quote_plus(response.locale.encode('utf-8')),
+        'PRODUCT': urllib.quote_plus(response.product.encode('utf-8')),
+        'VERSION': urllib.quote_plus(response.version.encode('utf-8')),
+        'PLATFORM': urllib.quote_plus(response.platform.encode('utf-8')),
+        'HAPPY': 'happy' if response.happy else 'sad',
+    }
+    return url.format(**data)
```
mythmon: Is `url` here considered safe, trusted data? If not, there are some DoS attacks when users can control format strings.

fjord/suggest/providers/trigger/provider.py
```diff
@@ -37,6 +39,20 @@ def build_ga_category(rule):
     return 'trigger_{rule.slug}_{rule.id}'.format(rule=rule)
 
 
+def interpolate_url(url, response):
+    if '{' not in url:
+        return url
```
mythmon: You do this early-out optimization twice. Once here, and once in `TriggerRedirect.handle_redirect`.

fjord/suggest/providers/trigger/provider.py
```diff
@@ -61,15 +77,25 @@ def handle_redirect(self, request, redirect):
...
             return
 
+        # Interpolate the url.
+        destination_url = rule.url
+        if '{' in destination_url:
+            # Fetch the response so we can interpolate the url.
+            try:
+                response = FeedbackResponse.objects.get(id=response_id)
+            except FeedbackResponse.DoesNotExist:
+                return
```
mythmon: Should there be a test for this case?

fjord/suggest/providers/trigger/provider.py
willkg: url is explicitly set by analyzers in the form. It can't come from anywhere else.

Mmm... I should switch this to a series of `.replace()` calls.

fjord/suggest/providers/trigger/provider.py
willkg: We could, but it seems pretty straight-forward and it's a very unlikely scenario. The only time feedback is deleted is if it's hate speech and I've noticed it.

fjord/suggest/providers/trigger/provider.py
willkg: They're two different optimizations, though. The one in `TriggerRedirect.handle_redirect` reduces the work involved--it doesn't pull the response and doesn't try interpolating. This one is here because this function isn't tied to `TriggerRule` and it's nice when functions can stand on their own because they're easier to use elsewhere. In this case we don't use it elsewhere.

Having said that, this should have a docstring.

willkg:
^^^ Rebases against master and also fixes `interpolate_url` to be safer and more documented.

mythmon:
That's all I had. The rest of this looks solid. r+

willkg:
Awesome--thank you!

PR #664, opened 2015-09-08
[bug 1202288] Update more requirements

willkg:
* chardet: 2.2.1 -> 2.3.0 (moved to vendor/src/, too)
* django-extensions: 1.3.3 -> 1.5.6
* django-pipeline: 1.5.2 -> 1.5.4
* tldextract: 1.3.1 -> 1.7.1 (moved to vendor/src/, too)


willkg:
tldextract affects computing the top-level domain from a url which is used when indexing.

I don't know offhand where chardet gets used.

django-pipeline had a minor version bump. It gets used when collecting static files during deployment and rendering templates. Any issues here would get sussed out by the unit test suite.

django-extensions provides a few developer-friendly `manage.py` commands.

To test this out, I:

1. ran `./manage.py collectstatic`
2. ran the unit tests
3. updated a vagrant vm and did the above there, too

There are no code changes here. If Travis is green, I'll merge and push to -stage and run smoketests there, then push to -prod.

PR #665, opened 2015-09-08
[bug 1202288] Move mock back to vendor/packages/

willkg:
We need to sdist version of mock since the sdist process makes changes
to packaging files. Without those changes, mock doesn't load.

This also adds the correct pyquery to vendor.pth which only affects my
local development environment.


willkg:
I'll wait for Travis to go green, then land this, push to -stage and continue testing there.

willkg:
Green -> mergine.

PR #670, opened 2015-09-09
[bug 1202288] Remove jingo-minify

willkg:
I don't know why it was still around--we don't use it anymore.


willkg:
I'll wait for Travis to turn green, then push it to -stage. This touches deploys, so that's the only good way to test it.

willkg:
Green! Merging!

PR #674, opened 2015-09-10
1203119 ratelimit

willkg:
r?

### DIFF
--- a/fjord/base/utils.py
+++ b/fjord/base/utils.py
@@ -18,7 +18,7 @@ from django.utils.encoding import force_str, force_text
 from django.utils.feedgenerator import Atom1Feed
 
 from product_details import product_details
-from ratelimit.helpers import is_ratelimited
+from ratelimit.utils import is_ratelimited
 from rest_framework.throttling import BaseThrottle
 from statsd.defaults.django import statsd
 
@@ -279,7 +279,7 @@ class Atom1FeedWithRelatedLinks(Atom1Feed):
                 attrs={'href': item['link_related'], 'rel': 'related'})
 
 
-def actual_ip(req):
+def actual_ip(group, req):
     """Returns the actual ip address
 
     Our dev, stage and prod servers are behind a reverse proxy, so the ip
@@ -295,7 +295,7 @@ def actual_ip(req):
 
 def actual_ip_plus_context(contextfun):
     """Returns a key function that adds md5 hashed context to the key"""
-    def _actual_ip_plus_context(req, *args, **kwargs):
+    def _actual_ip_plus_context(group, req, *args, **kwargs):
         # Force whatever comes out of contextfun to be bytes.
         context = force_str(contextfun(req))
 
@@ -305,7 +305,7 @@ def actual_ip_plus_context(contextfun):
         context = hasher.hexdigest()
 
         # Then return the ip address plus a : plus the desc md5 hash.
-        return actual_ip(req) + ':' + context
+        return actual_ip(group, req) + ':' + context
     return _actual_ip_plus_context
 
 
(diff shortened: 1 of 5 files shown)


willkg:
@mythmon Can you look through this to make sure I didn't mess something up?

fjord/base/utils.py
```diff
@@ -279,7 +279,7 @@ def add_item_elements(self, handler, item):
                 attrs={'href': item['link_related'], 'rel': 'related'})
 
 
-def actual_ip(req):
+def actual_ip(group, req):
```
willkg: django-ratelimit changed the function signature for key functions from `(request)` to `(group, request)`.

fjord/base/utils.py
```diff
@@ -323,14 +323,30 @@ def ratelimit(rulename, keyfun=actual_ip, rate='5/m'):
...
+        must check the request for 'limited' and redirect/handle
+        accordingly.
+
+        For example::
+
+            @ratelimit(rulename='foo', keyfun=actual_ip, rate='1/10m')
+            def view(request):
+                if getattr(request, 'limited', False):
+                    # handle ratelimit here!
+                    raise PermissionDenied()
```
willkg: I added this note so I don't have to go figure out what's going on again. This is one of the big differences between our `ratelimit` decorator and the django-ratelimit one.

fjord/base/utils.py
```diff
@@ -323,14 +323,30 @@ def ratelimit(rulename, keyfun=actual_ip, rate='5/m'):
...
     def decorator(fn):
         @wraps(fn)
         def _wrapped(request, *args, **kwargs):
             already_limited = getattr(request, 'limited', False)
             ratelimited = is_ratelimited(
-                request=request, increment=True, ip=False, method=['POST'],
-                field=None, rate=rate, keys=keyfun)
+                request=request, group=rulename,
+                key=keyfun, rate=rate, method=['POST'],
+                increment=True)
```
willkg: Documentation for the new `is_ratelimited`: http://django-ratelimit.readthedocs.org/en/latest/usage.html#is_ratelimited

We use the rulename as the group since that's essentially what rulename is. If we didn't pass in rulename as the group value, then it'd generate a group based on the dotted name of the function. That doesn't work here and would create problems.

fjord/base/utils.py
```diff
@@ -387,13 +403,15 @@ def parse_rate(self, rate):
...
+            request=request, group=self.rulename,
+            key=self.keyfun, rate=self.rate, method=self.methods,
+            increment=True)
 
         if ratelimited:
-            # Failed rate-limiting, so this request is not allowed.
-            statsd.incr('throttled.' + self.rulename)
+            if not already_limited:
+                statsd.incr('throttled.' + self.rulename)
             return self.throttle_failure()
```
willkg: I added the `already_limited` thing because otherwise in theory a request for a view with multiple throttles that all fail will do multiple `statsd.incr` calls. In practice, it might be moot since django rest framework might halt things at the first throttle that fails. I don't really know and don't know if it'll change at some point in the future.

fjord/feedback/tests/test_api.py
```diff
@@ -926,10 +924,12 @@ def data_generator():
...
             # django-ratelimit fails the throttling if we hit the url
-            # a fajillion times in rapid succession. For now, we add
-            # a sleep which means this test takes 5 seconds now.
+            # a fajillion times in rapid succession. For now, we add a
+            # sleep which makes this test run a little longer than it
+            # probably needs to.
+            #
             # FIXME: Look into this more for a better solution.
-            time.sleep(0.05)
+            time.sleep(0.01)
```
willkg: 0.01 works locally. If I remove this altogether, it fails. I didn't want to take the time to look into it. If it fails on travis, I'll increase this again.

mythmon:
To summarize these changes (to make sure I understand):

* Add a `group` parameter to all the key functions (which is generally ignored by Fjord stuff)
* Make the tests run faster by removing a speed up loop.
* Add some docstrings.

Is that accurate? If so, this looks pretty straight forward, r+.

I'd really like to know why that sleep is needed though. Maybe some day I'll find out.

willkg:
The commit messages should cover the summary of what happened:

https://github.com/willkg/fjord/commit/d00b49f37016c5ce061c621fc99b7b2d87ef11ea

> This tweaks the ratelimit for incoming feedback and the throttle test
> they both use a "constant".
> 
> It also cleans up some pep8 issues.

https://github.com/willkg/fjord/commit/7e23467bccb4afb9138620ed713114fa58987ecf

> This updates django-ratelimit to the latest version. There were some big
> changes in django-ratelimit, so this updates the fjord code to account
> for those changes.
> 
> * ratelimit and is_ratelimited function signature changes
> * key function takes (group, request) as args now
> 
> I also reduced the sleep we use to make the throttle test work from 0.05
> to 0.01. That works locally, but we'll have to see if that works on
> travis, too.

I have no idea why that sleep is needed--it's puzzling.

willkg:
I think the summary is close enough. Merging this now.

Thank you!

PR #676, opened 2015-09-10
1203173 analyzer tests

willkg:
r?

### DIFF
--- a/fjord/analytics/analyzer_views.py
+++ b/fjord/analytics/analyzer_views.py
@@ -15,23 +15,17 @@
 # English. Ergo, there's no localization here.
 
 from datetime import date, datetime, timedelta
-from pprint import pformat
 import csv
 
 from elasticsearch_dsl import F
 
-from django.core.paginator import Paginator, PageNotAnInteger, EmptyPage
-from django.core.urlresolvers import reverse_lazy
 from django.http import HttpResponse
 from django.shortcuts import get_object_or_404, render
 from django.utils.encoding import force_bytes
-from django.views.generic.edit import FormView, CreateView, UpdateView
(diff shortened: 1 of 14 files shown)


willkg:
This couples a some code reorganization and code cleanup with addition of tests for analyzer views that didn't have any.

These are basic tests--they mostly test to see if the view "works" and the template "rendered" for the basic situations. They don't test edge cases or complex functionality.

There's still a lot of goofy code that should be rewritten in there. I'm going to push that off to another day.

@mythmon Can you take a look at these?

mythmon:
I didn't look at the views that moved. I assume those are just moves and there weren't any edits. If that's not the case, I can look harder.

The tests look straight forward. Most of the tests seem to have been copy-pasted, which is fine, but it means they have the same oddities. `jane@example.com` is one, and generally having multiple logically independent things in one test is another.

Yay for more tests!

willkg:
That's correct--the views just moved from one place to another. I thought about rewriting them as CBVs, but decided to push that off.

I'll remove the `user__email` stuff in a cleanup commit. At some point, I'll wrap my head around pytest fixtures and use that. Maybe I'll do that now ... Mmm....

willkg:
Oh, wait--I remember: pytest fixtures don't work well with Django TestCase derivatives. So I won't do that.

willkg:
Bah! I'm mis-remembering: fixtures are fine, it's paramaterize that doesn't work so well.

willkg:
Tried to get fixtures to work, but couldn't and decided not to spend more time on it.

Did some minor cleanup. If Travis is green, I'll merge it.

willkg:
Green. Merging!

PR #677, opened 2015-09-11
Cosmetic: Change relative imports to absolute imports

willkg:
Most of the code base uses absolute imports, but we used relative imports in a few places. The case for using relative imports is that it theoretically makes it easier to pluck the code and stick it somewhere else. Sure, but there's no way we're moving this code somewhere else, so I'd rather the imports were clearer and more explicit.

Given that, I changed all the relative imports to absolute imports.

Minor r?

### DIFF
--- a/fjord/__init__.py
+++ b/fjord/__init__.py
@@ -13,6 +13,7 @@ _has_setup_vendor_path = False
 
 ROOT = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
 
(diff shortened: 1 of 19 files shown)


willkg:
Actually, I think if travis is cool with this, I'll just merge it. It's pretty straight-forward. Bad imports will fail immediately.

willkg:
Travis is green--merging!

PR #678, opened 2015-09-14
[bug 1202647] Update dev requirements

lgp171188:
Update the dependencies listed in the bug 1202647.
Get rid of flake8 and its dependencies since we no longer use it
directly. pre-commit installs flake8 from its dependencies in its
own virtualenv and does not use the one we have been installing till
now.


willkg:
Looks great! Thank you!

PR #679, opened 2015-09-14
Cosmetic: Clean up extract, merge and l10n modules

willkg:
* pep8 fixes
* removed some of the code needed for older versions of Babel
* renamed some things to reduce "tower" references


willkg:
This is all in the developer infrastructure and it's cosmetic code cleanup, so I'm going to wait for Travis to greenlight it and then land it.

PR #680, opened 2015-09-15
[bug 1205013] Add comprehensive test plan

willkg:
Periodically, we do substantive changes that affect the whole codebase.
We have a unit test suite and a set of smoketests, but there's a lot of
stuff that isn't covered by those.

Instead of writing up a test plan every time we do a substantive change,
this is an attempt at writing a comprehensive test plan covering all the
important things that we can reuse every time we do a substantive
change.

r?

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -212,6 +212,14 @@ Unit tests
 1. run ``./py.test`` to run the unit tests with a clean database
 
 
+JavaScript tests
+----------------
+
+1. run ``./manage.py runserver``
+2. open your browser
+3. go to http://127.0.0.1:8000/static/tests/index.html
+
+
 Smoketests
 ----------
 
@@ -249,7 +257,7 @@ Testing Elasticsearch and indexing
    index some stuff
 3. run ``./manage.py esstatus`` to make sure the new index is there
 4. run ``./manage.py esdelete <index>`` to delete that index
-5. run ``./manage.py esstatus`` to make sure the index was created
+5. run ``./manage.py esstatus`` to make sure the index was deleted
 6. run ``./manage.py esreindex --percent=5`` to recreate the index
 7. verify feedback is indexed
 
@@ -340,7 +348,7 @@ With an existing vagrant environment:
 
 1. run ``vagrant up``
 2. run ``./peep.sh install -r requirements/requirements.txt``
-3. run ``./peep.sh install -r requirements.dev.txt``
+3. run ``./peep.sh install -r requirements/dev.txt``
 4. run ``vagrant ssh``
 
    1. run ``cd fjord``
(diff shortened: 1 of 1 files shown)


willkg:
@mythmon Can you skim these changes? Can you think of other things we should add to this test plan?

docs/testing.rst
```diff
@@ -167,19 +170,191 @@ the relevant place and then create a new ``test_FILENAMEHERE.js`` file
...
+
+Testing Elasticsearch and indexing
+----------------------------------
+
+1. run ``./manage.py esstatus``
+2. run ``./manage.py esreindex --percent=5`` to create a new index and
+   index some stuff
+3. run ``./manage.py esstatus`` to make sure the new index is there
+4. run ``./manage.py esdelete <index>`` to delete that index
+5. run ``./manage.py esstatus`` to make sure the index was created
```
mythmon: Should this say "to make sure the index was deleted"?

docs/testing.rst
```diff
@@ -167,19 +170,191 @@ the relevant place and then create a new ``test_FILENAMEHERE.js`` file
...
+4. run ``vagrant ssh``
 
-On those rare, wonderful occasions when we get to remove code, we
-should remove the tests for it, as well.
+   1. run ``cd fjord``
+   2. run ``./py.test``
 
-If we liberate some functionality into a new package, the tests for
-that functionality should move to that package, too.
+If that works, then it probably works fine in vagrant development
```
mythmon: > If that works, then it probably works fine in vagrant development environment.

I don't understand this sentence. Do you mean that if the tests work in the Vagrant environment, it is likely that everything else does too?

docs/testing.rst
```diff
@@ -167,19 +170,191 @@ the relevant place and then create a new ``test_FILENAMEHERE.js`` file
...
+
+
+Testing vagrant
+---------------
+
+With an existing vagrant environment:
+
+1. run ``vagrant up``
+2. run ``./peep.sh install -r requirements/requirements.txt``
+3. run ``./peep.sh install -r requirements.dev.txt``
```
mythmon: I think this should be `requirements/dev.txt`.

mythmon:
Does your test plan include running the JS tests? Or do you assume the manual tests will be good enough there? Or do you have something like Sumo's selenium-qunit tester?

I like all these explicit instructions. I think it covers a lot of cases. It is really good to write this down.

In the future, I imagine a lot of these could be made into a script and/or done as a part of CI, which would be cool.

docs/testing.rst
willkg: Oops!

docs/testing.rst
willkg: Yes. I'll fix that sentence--it's sloppy.

docs/testing.rst
willkg: Oops!

willkg:
Oops--forgot about the JS tests. I'll add that.

It definitely makes sense to automate more of this. More automation and less manual testing is definitely good, but I think we'll always need some manual testing.

willkg:
@mythmon Thank you for reviewing!

PR #681, opened 2015-09-16
[bug 1205370] Add smoketest for qunit tests

willkg:
I lifted this code from Kitsune and modified it slightly so it works in the Fjord smoketest system.

Quick r?


mythmon:
I set up the smoke tests and ran them. This test ran, and it passed. This approach worked fine on Sumo for a while, so this all looks good.

r+

willkg:
Thank you!

PR #682, opened 2015-09-17
[bug 1146686] Remove django-cronjobs

willkg:
Input doesn't use django-cronjobs. This removes it from the requirements
and vendor/.


willkg:
This isn't cosmetic, but it's not used, so it shouldn't affect anything. I'll wait until Travis is green, then merge it.

PR #684, opened 2015-09-23
1146686 django18

willkg:
This updates Fjord to Django 1.8 and django-jinja. Long live the new Fjord!

r?


willkg:
I wrote up a comprehensive testing plan hre:

http://fjord.readthedocs.org/en/latest/testing.html#comprehensive-test-plan

That'll help here figuring out whether things are good. I also wrote up a draft blog post covering updating to Django 1.8 and django-jinja here:

http://bluesock.org/~willkg/blog/mozilla/input_django_1_8_upgrade.html

That covers some of what happened, how and why. I passed that draft post around and everyone seemed ok with it. I claim that suggests that the bulk of the decisions made are good enough.

@mythmon Can you look this over? You can skip all the .html file changes--those are mostly file moves with some minor tweaks which will throw errors in the tests if there were typos.

fjord/base/jinja2/includes/common.html
```diff
@@ -7,7 +7,7 @@
         <label for="language">{{ _('Other languages:') }}</label>
         <select id="language" name="lang" dir="ltr">
           {% for code, name in LANGUAGES|dictsort -%}
-            <option value="{{ code }}" {{ code|ifeq(LANG|lower, "selected") }}>
+            <option value="{{ code }}" {{ "selected" if code == LANG.lower() }}>
```
mythmon: Wow. You're write, `|ifeq` is so weird.

fjord/base/management/commands/extract.py
```diff
@@ -15,48 +14,29 @@
...
-
-
-TOWER_KEYWORDS = dict(DEFAULT_KEYWORDS)
-
-
-if hasattr(settings, 'TOWER_KEYWORDS'):
-    TOWER_KEYWORDS.update(settings.TOWER_KEYWORDS)
+TEXT_DOMAIN = 'django'
+KEYWORDS = dict(DEFAULT_KEYWORDS)
+KEYWORDS.update({'_lazy': None})
```
mythmon: That seems odd. Is there any particular reason to do this instead of `KEYWORDS['_lazy'] = None`? Is it just similar to the line `TOWER_KEYWORDS.update(settings.TOWER_KEYWORDS)` that used to be there?

willkg: It's possible I had more stuff in there, but then nixed it later. I'm not really sure. I don't see anything in my notes. I'll change it since the `KEYWORDS['_lazy'] = None` is less curious looking.

fjord/base/jinja2/includes/common.html
willkg: It seems unnatural and unreadable to me Maybe there was a use case it was good at?

willkg:
^^^ Rebases against master. Fixes a line to make it less curious (cosmetic).

fjord/manage_utils.py
```diff
@@ -172,25 +173,54 @@ def monkeypatch():
...
+
+        * it does *not* capture all the Jinja2 templates used to render.
+        Only the topmost one requested by the render() function.
+
+        """
+        @wraps(fun)
+        def _more_info(request, template_name, *args, **kwargs):
+            resp = fun(request, template_name, *args, **kwargs)
+
+            # If we're in a test situation, then capture it all
```
mythmon: The comment here implies that the context is only captured during tests, but it seems it is always captured?

mythmon:
whew. I read through it all, I don't see anything else to comment on.

The tests pass. The smoketest pass. I went through most of the testing plan both with and without this patch, and they behaved the same. I skipped over

* Anything that required persona log in, because I can't get that to work.
* Cron jobs, because there are a lot of those and I don't know what most of them are supposed to do and I'm lazy.
* CELERY_ALWAYS_EAGER = False. `./manage.py celeryd` didn't work for me.
* Localizations, because I don't have those.

I don't think this PR would change the results of those tests. I think the test plan doesn't test the trigger system. So I didn't do that either.

I say this looks good, and it seems to work. :ship:.

fjord/manage_utils.py
willkg: Umm... Yeah. So this code doesn't affect whether it kicks on or not, but rather the code here:

https://github.com/mozilla/fjord/pull/684/files#diff-77b722d1cd2658a4eda87c2b19b62ca3R27

I'll nix this comment since it seems weird here. I'll add a docstring to `monkeypatch_render` that says only to use it for test contexts.

willkg:
^^^ Rebased against master again. Clarified the `monkeypatch_render` comment and docstring.

I think that's good. When Travis is green, I'll push this to -stage and see how it all goes.

willkg:
I pushed this to stage and it seems fine. I'm going to land it and push it to stage

willkg:
Thank you!

PR #685, opened 2015-09-29
[bug 1209787] Fix tests sending GA data

willkg:
We have a handful of google_utils tests that need to test that function.
This fixes them so they're all mocked and none of them send actual data.

We have a bunch of other tests that test code that calls google_utils
functions. This fixes the google_utils functions to not send any data
when running in a test context.

r?

### DIFF
--- a/fjord/base/google_utils.py
+++ b/fjord/base/google_utils.py
@@ -31,6 +31,9 @@ def ga_track_event(params, async=True):
        testing/development data hosing your production data.
 
     """
+    if sys.argv and sys.argv[0].endswith('py.test'):
+        return
+
     params.update({
         'v': '1',
         # FIXME: turn this into a setting. the complexity is that it's
@@ -41,7 +44,7 @@ def ga_track_event(params, async=True):
 
     # If we're in DEBUG mode or running tests, we don't want to hose
     # production data, so we prepend test_ to the category.
-    if 'ec' in params and (settings.DEBUG or 'test' in sys.argv):
+    if 'ec' in params and settings.DEBUG:
         params['ec'] = 'test_' + params['ec']
 
     if async:
--- a/fjord/base/google_utils.py
+++ b/fjord/base/google_utils.py
@@ -31,6 +31,12 @@ def ga_track_event(params, async=True):
        testing/development data hosing your production data.
 
     """
+    # FIXME: We should only have the google id in settings in
+    # production and not anywhere else. Thus when not running in
+    # production, this function would check that setting, notice it
+    # was blank and then exit. Then we don't have to do this goofy
+    # "are we testing?" test which breaks when we change our testing
+    # infrastructure.
     if sys.argv and sys.argv[0].endswith('py.test'):
         return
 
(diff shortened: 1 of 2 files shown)


willkg:
@mythmon If Travis is good with this, can you review it?

fjord/base/tests/test_google_utils.py
```diff
@@ -77,24 +80,32 @@ def test_ga_track_event_non_test(self):
...
-            d = 'fjord.base.google_utils.post_event.delay'
-            with patch(d) as delay_patch:
-                params = {'cid': 'xxx', 'ec': 'ou812'}
-                ga_track_event(params, async=True)
-                assert delay_patch.call_count == 1
+            post_event = 'fjord.base.google_utils.post_event'
+            post_event_delay = 'fjord.base.google_utils.post_event.delay'
+            with patch(post_event) as post_event_patch:
+                with patch(post_event_delay) as post_event_delay_patch:
+                    params = {'cid': 'xxx', 'ec': 'ou812'}
```
mythmon: Not a problem, but in python 2.7 and above (I think), you could do

```python
with patch(post_event) as post_event_patch, patch(post_event_delay) as post_event_delay_patch:
    params = ...
```

That might blow your line length limit though, which would sort of ruin the point. *shrug*

fjord/base/tests/test_google_utils.py
```diff
@@ -33,8 +35,9 @@ def handler(*args, **kwargs):
 
 
 class TestGoogleUtils(TestCase):
-    @set_sys_argv(['test'])
-    def test_ga_track_event_test(self):
+    @override_settings(DEBUG=True)
+    @set_sys_argv([])
```
mythmon: This is... weird. I can't think of a better way to do it, but munging `sys.argv` seems nasty.

fjord/base/google_utils.py
```diff
@@ -31,6 +31,9 @@ def ga_track_event(params, async=True):
        testing/development data hosing your production data.
 
     """
+    if sys.argv and sys.argv[0].endswith('py.test'):
```
mythmon: Acting differently while under test seems like a good way to make hard to reason about tests. Is there another way to deal with this?

mythmon:
This is icky, but I think the amount of icky remained the same before and after this patch. It is just updating the do-not-run-in-tests guard to understand py.test. Given that, I guess this is fine.

fjord/base/google_utils.py
willkg: Yeah... The other way to do it is like what we do with gengo where we only have the real credentials on production and the other environments don't have them. Then if you don't have them, then it no workey.

However, GA is slightly different in that our id is both in a js file and in python. So we could do a setting, but we'd need to do it in a way that it's obvious it only affects the python code or something like that.

I'm not sure what other options there are.

fjord/base/tests/test_google_utils.py
willkg: It's essentially the same thing as overriding Django settings, but it's at a system level. Because it's at the system level, it could have other unintentional effects so it's less desirable.

Probably best to add a warning to the `set_sys_argv` docstring about that.

YOLO!

fjord/base/tests/test_google_utils.py
willkg: Hot shit! I had no idea! Maybe you could use parens, too?

willkg: Can't use parens. But can use `\`.

willkg:
^^^ That addresses everything except the nested `with`. I redid that with `\`, but then the linter complained and I decided my patience for fiddling with things to make tools happy today had run out.

When Travis says green, I'll land it.

Thank you!

PR #686, opened 2015-09-30
[bug 1146686] Fix triggerrule testing

willkg:
The triggerrule testing thing was sending [''] for fields like locale
which are \n delimited textareas. That now causes an HTTP 400 error.

This fixes that by filtering out empty strings before generating the
request to send to the API.

Note: This commit was buried in the django-1.8 pr. I cherry-picked it and removed it from the Django 1.8 pr.

r?

### DIFF
--- a/fjord/suggest/providers/trigger/static/js/triggerrule.js
+++ b/fjord/suggest/providers/trigger/static/js/triggerrule.js
@@ -94,15 +94,26 @@
         }
 
         /**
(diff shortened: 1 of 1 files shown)


willkg:
This isn't covered by the unit tests. The smoketests don't cover the analyzer stuff, either. I re-tested it by hand and verified that it's broken in master and works with this patch.

PR #688, opened 2015-10-01
[bug 1200369] Update elasticsearch-dsl to 0.0.8

willkg:
NB: The 0.0.8 on pypi has no related tag in the elasticsearch-dsl
repository, so I picked a commit that looks like it might be 0.0.8, but
I don't know for sure.


willkg:
Once Travis is green, I'll land this.

PR #689, opened 2015-10-02
Update dennis to 0.7

willkg:


willkg:
If Travis is cool with this, I'll land it.

PR #690, opened 2015-10-02
[bug 1205296] Remove django-mobility

willkg:
We were only using mobile_template, so I grabbed that and then ditched
django-mobility.


willkg:
If travis is cool, then we can land this. It'll either fail loudly or be totally fine.

PR #692, opened 2015-10-05
Properly name the vagrant VM

lgp171188:
Name the vagrant VM as 'fjord-dev' instead of letting vagrant
name it as fjord_default_<some long random number>.


willkg:
Thank you for this! I had no idea you could do that.

PR #693, opened 2015-10-10
[bug 1213340] Add fields to querystring in hb data view

willkg:
Gregg said it helps to be able to restrict the hb data list to just
fields he's interested in. This helps him debug his
problem-of-the-moment.

I wanted to implement this quickly, so I added a fields param to the
querystring as well as some minor stuff to make sure it's persisted via
links. I also added a list of all the fields at the bottom of the page
so Gregg can concoct his urls by hand.

If this turns out to be much used, we'll want to add a form to make it
easier to select and deselect fields.

r?


willkg:
@mythmon Does this look good enough?

fjord/heartbeat/analyzer_views.py
```diff
@@ -19,38 +19,65 @@ def hb_data(request, answerid=None):
...
+    showdata_filter = None
 
     if answerid is not None:
         answer = Answer.objects.get(id=answerid)
 
     else:
+        # Get fields and split on comma, then get rid of empty fields.
+        fields = request.GET.get('fields', '').split(',')
+
+        # Make sure all the fields
```
mythmon: I think you left off halfway through this sentence.

willkg: Oops!

Hey, it's Saturday--you shouldn't be looking at this. Go outside and play!

willkg:
^^^ Fixes the unfinished comment.

fjord/heartbeat/analyzer_views.py
```diff
@@ -19,38 +19,65 @@ def hb_data(request, answerid=None):
...
 
     if answerid is not None:
         answer = Answer.objects.get(id=answerid)
 
     else:
+        # Get fields and split on comma, then get rid of empty fields.
+        fields = request.GET.get('fields', '').split(',')
+
+        # Nix any fields that don't exist
+        fields = [field for field in fields if field and field in ALL_FIELDS]
```
mythmon: Nothing to fix here, this line is just really funny.

mythmon:
This looks good. The only thought I have is that the UI code you wrote looks like the kind of thing Django Forms are supposed to "fix", specifically the widget system. Though I don't actually have much experience with Django Forms, so I'm not sure if it is actually similar.

r+

fjord/heartbeat/analyzer_views.py
willkg: Mmm... I can drop the "if field" part because if it's falsey, it won't be in `ALL_FIELDS`. I'll do that.

willkg:
Yeah... Mostly I'm dragging my feet because I don't want a big form on the page and don't know of a better way to do it. I did think about doing it as a text-box, but ... that's kind of like fiddling with the querystring.

Anyhow, I think it's worth pushing off that work until I have a better feel for how Gregg will use this.

willkg:
Thank you!

PR #694, opened 2015-10-14
[bug 1214272] Rework analyzer products views

willkg:


willkg:
I'll skim the code, then wait for travis to go green, then land it. This only affects analyzers.

willkg:
Green! Merging!

PR #696, opened 2015-10-20
[bug 1216624] Add --check-header to msgfmt run

willkg:
This will cause msgfmt to check the headers of the .po file before it
compiles it to a .mo file. In most cases, the header values don't matter
a lot. However, this validates the Plural-Forms header which is important.

If Plural-Forms header is invalid, then msgfmt will exit with a status
code of 1 and we don't replace the existing .mo file. If it's fine,
then we copy the targettmp file to the destination and we're good to go.

r?

### DIFF
--- a/bin/compile-linted-mo.sh
+++ b/bin/compile-linted-mo.sh
@@ -22,6 +22,8 @@ function compilemo() {
     then
         mv "${targettmp}" "${target}"
     fi
+
+    rm -rf "${targettmp}"
 }
 
 # Get the path to the python binary from the "-p <PYTHON>"
--- a/bin/compile-linted-mo.sh
+++ b/bin/compile-linted-mo.sh
@@ -8,8 +8,20 @@ function compilemo() {
     dir=`dirname $1`
     stem=`basename $1 .po`
 
+    target="${dir}/${stem}.mo"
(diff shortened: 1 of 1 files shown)


willkg:
@mythmon How does this look to you?

The end result are these two scenarios:

1. Language XX is added to Fjord and has never been compiled to a `.mo` file. Turns out it has an invalid Plural-Forms header. `compile-linted-mo.sh` kicks up an error and does not generate a `.mo` file.
2. Language YY has been part of Fjord for a while and has been compiled to a `.mo` file. Someone accidentally messes up the Plural-Forms header and it's now invalid. `compile-linted-mo.sh` kicks up an error and does not generate a new `.mo` file--the old one is left in place.

Does that sound ok? Does the code match that behavior?

mythmon:
Both those scenarios sound reasonable, and this seems to match what we do with Dennis lint errors too. From a localizers point of view, this is a modification to the l10n linting process so that it catches more errors.

The code looks right. I think you should probably `rm -f "${targettmp}"` in the case of a failure (in an else block), just to keep things tidy.

willkg:
^^^ That deletes the tmp file.

Merging this now. Thank you!

PR #697, opened 2015-10-20
WIP fiddling with whitespace -- don't judge me! WIP

willkg:


fjord/base/l10n.py
```diff
@@ -42,54 +41,37 @@ def __init__(self, environment):
...
+
+            {% trans foo=bar, trimmed %} ... {% endtrans %}
+
+        """
+        parse_block = InternationalizationExtension._parse_block
+        ref, buffer = parse_block(self, parser, allow_pluralize)
+
+        if self._yolo_is_trimmed():
+            buffer = collapse_whitespace(buffer)
+        return ref, buffer
```
willkg: This new `_parse_block` only collapses whitespace if there's a "trimmed" in the thingy. For Django blocktrans tags, that "thingy" is a flag indicating that the whitespace should be collapsed. Here, it's a variable and we're (ab)using inspect and stackframes to see whether it's there and collapsing accordlingly.

The nice thing this affords us is this:

1. You can use tower's extract (sans all the "collapse all the whitespace!") command. It uses babel which compiles the Jinja2 template to AST and then walks the AST for trans nodes. Since we're collapsing the whitespace in the block, it picks up the text and that's the msgid. This works when rendering the template, too. Everything is rainbows.
2. You can use django-jinja's makemessages which overrides django's makemessages. It's looking for things that look like django's stuff, so it sees the "trimmed" and it collapses the whitespace. That's "good enough".

Further, this lets us specify which blocks should be trimmed. I'm not sure why you would want to do otherwise. I kind of think this is a stupid default. But this is closer to the Django behavior.

fjord/feedback/jinja2/feedback/generic_feedback.html
```diff
@@ -40,7 +40,7 @@
                 </p>
               </noscript>
 
-              {% trans product=product.display_name %}
+              {% trans product=product.display_name, trimmed %}
```
willkg: This is how we'd have to do all trans blocks if we landed this.

willkg:
Why do this?

1. it's closer to what vanilla Django does.
2. it allows us to have trans blocks that don't collapse whitespace. that might be helpful for sending text-based emails? i don't know.
3. less code to maintain, though we're trading that for stack inspection.

PR #699, opened 2015-10-22
[bug 1216998] Move all l10n under fjord.l10n

willkg:
* create a fjord.l10n app to hold all l10n stuff
* move extract and merge commands to l10n app
* move fjord/base/l10n.py to fjord/l10n/ext.py and clean up
* add tests for collapse_whitespace and TowerI18nExtension
* add basic tests for extract and merge that mostly test to see if the
  commands run at all


fjord/settings/base.py
```diff
@@ -581,10 +582,12 @@ def lazy_langs():
...
-        ('fjord/**.py', 'fjord.base.l10n.extract_python'),
-        ('fjord/**/templates/**.html', 'fjord.base.l10n.extract_template'),
-        ('fjord/**/jinja2/**.html',  'fjord.base.l10n.extract_template'),
-        ('templates/**.html', 'fjord.base.l10n.extract_template'),
+        ('fjord/**.py', 'babel.messages.extract.extract_python'),
+        ('fjord/**/jinja2/**.html',  'jinja2.ext.babel_extract'),
+        # FIXME: These are using Jinja2 template extraction for Django
+        # templates. Does that work right?
+        ('fjord/**/templates/**.html', 'jinja2.ext.babel_extract'),
+        ('templates/**.html', 'jinja2.ext.babel_extract'),
```
willkg: We should provide helpers for these because otherwise they're hard to remember. Maybe something like:

* `fjord.l10n.extractors.extract_python`
* `fjord.l10n.extractors.extract_jinja2`
* `fjord.l10n.extractors.extract_django`

fjord/l10n/management/commands/extract.py
```diff
@@ -49,21 +57,82 @@ def create_pofile_from_babel(extracted):
...
+    2. Jinja2's ``babel_extract`` stops adding
+       InternationalizationExtension to the environment if it's
+       not there
+
+    """
+    import jinja2.ext
+    from fjord.l10n.ext import TowerI18nExtension
+
+    jinja2.ext.InternationalizationExtension = TowerI18nExtension
+    jinja2.ext.i18n = TowerI18nExtension
```
willkg: This is so cray-cray.

fjord/l10n/management/commands/extract.py
```diff
@@ -49,21 +57,82 @@ def create_pofile_from_babel(extracted):
...
+    help = 'Extracts strings for translation.'
+
     def add_arguments(self, parser):
         parser.add_argument(
             '--domain', '-d', default=DEFAULT_DOMAIN, dest='domain',
             help=(
                 'The domain of the message files.  If "all" '
                 'everything will be extracted and combined into '
-                '%s.pot. (default: %%default).' % TEXT_DOMAIN
+                '%s.pot. (default: %%%%default).' % TEXT_DOMAIN
```
willkg: This gets double-unescaped to become just `%default` which gets replaced with the default value.

fjord/l10n/management/commands/extract.py
```diff
@@ -41,6 +44,11 @@ def create_pounit(filename, lineno, message, comments):
         unit.addnote(comment, 'developer')
 
     unit.addlocation('%s:%s' % (filename, lineno))
+    # FIXME: Add variable format flags
+    # if python-format:
+    #     unit.settypecomment('python-format', present=True)
+    # if python-brace-format:
+    #     unit.settypecomment('python-brace-format', present=True)
```
willkg: xgettext (and thus "./mange.py makemessages") adds variable format flags. "./manage.py extract" does not so far.

I have no idea if any of the translations tools we have make use of them, but it'd let me fix some issues with dennis that I'm grappling with now if dennis knew exactly what variable formats were used in a given string.

fjord/l10n/management/commands/extract.py
```diff
@@ -49,21 +57,82 @@ def create_pofile_from_babel(extracted):
...
+    set up the environment, it checks to see if the environment
+    has InternationalizationExtension in it and if not, adds it.
+
+    https://github.com/mitsuhiko/jinja2/blob/2.8/jinja2/ext.py#L587
+
+    That stomps on our TowerI18nExtension so trans blocks don't get
+    whitespace collapsed and we end up with msgids that are different
+    between extraction and rendering. Argh!
+
+    Two possible ways to deal with this:
```
mythmon: Nit: you say `Two possible ways` and then list three.

fjord/l10n/utils.py
```diff
@@ -0,0 +1,9 @@
+import re
+
+
+_COLLAPSE_RE = re.compile(r'\s+', re.UNICODE)
+
+
+def collapse_whitespace(message):
+    """Collapses consecutive whitespace into a single space"""
+    return _COLLAPSE_RE.sub(' ', message).strip()
```
robhudson: `return u' '.join(map(lambda s: s.strip(), message.strip().splitlines()))` seems slightly faster and equivalent, but do your own testing.

```
 ~ python -mtimeit -s "message = u'  tést \n\t foo \nyo'" "u' '.join(map(lambda s: s.strip(), message.strip().splitlines()))"
1000000 loops, best of 3: 1.64 usec per loop
 ~ python -mtimeit -s "import re; message = u'  tést \n\t foo \nyo'; _COLLAPSE_RE = re.compile(r'\s+', re.UNICODE)" "_COLLAPSE_RE.sub(' ', message).strip()"
1000000 loops, best of 3: 1.96 usec per loop
```

willkg: Out of curiosity, how well does that work in the single line and (say) 10-line cases?

robhudson: For single line the `re` version is slightly faster: 1.02 usec per loop vs 0.768 (for re)
For 10 lines the `splitlines` is significantly faster: 3.95 usec per loop vs 11.5 (for re)

fjord/l10n/tests/test_ext.py
```diff
@@ -0,0 +1,162 @@
...
+    def test_trans_interpolation(self):
+        tmpl = (
+            '{% trans tag="<p>" %}\n'
+            '  this is a tag <b>{{ tag }}</b>\n'
+            '{% endtrans %}'
+        )
+        assert (
+            render(tmpl, LANGUAGE='en') ==
+            'this is a tag <b>&lt;p&gt;</b>'
+        )
```
willkg: Redid the extension tests so they're way easier to read/maintain. Plus they have a bunch of interpolation tests covering escaping.

fjord/l10n/utils.py
willkg: That's pretty interesting. Django uses a regexp:

https://github.com/django/django/blob/master/django/utils/translation/__init__.py#L243

I like that your version is faster for the important/likely scenarios. My one gripe is that it's less readable to me, but that's pretty minor. I want to test it with Python 3, but then I'd be game for changing it.

willkg:
Things left to do:

1. [ ] move settings to central place with sane defaults
2. [ ] reinstate some of the settings I removed because I didn't need them in fjord
3. [ ] look into what happened to the language activation code--do we need it anymore?

After that, I'll un-WIP this and land it. Then I'll pull it all into the tower-fork and additionally do the following:

1. [ ] write tests for extract command
2. [ ] write tests for merge command
3. [ ] write a compilemessages command--it's silly we all have bash scripts for this

fjord/l10n/utils.py
robhudson: fwiw, "my version" is what the jinja2-django-tags project does and where I saw it and got interested about the differences.
https://github.com/MoritzS/jinja2-django-tags/blob/master/jdj_tags/extensions.py#L188

willkg: Very interesting!

willkg:
I remember now. We ditched all the translation activation stuff because we switched to using `django.po` as the `.po` file. If we allowed for other filenames, then we'd need to override the activation code to allow for that. So... figuring out activation stuff--check!

willkg:
I think I've finished up all the work I need to do in fjord and it's ready to be moved to the tower-fork.

One key part of the docs will be "how do i migrate from old tower to tower-fork?" I'm keeping that in mind as I do things.

willkg:
Need to do a pass on the code, fix whatever that turns up and rewrite the `collapse_whitespace` code. Then this will be ready for review.

willkg:
@mythmon This is ready for reviewing. There are still some FIXME comments in there, but most of them don't affect Fjord. When we move the code to tower-fork, we can address them there.

The other caveat is that there's like zero testing for extract/merge. Those commands are really hard to test inside fjord without dealing with fjord's strings. They should be easier to test when extracted into a library.

willkg:
Scratch that. I decided to put the whole thing into a library. I'll do another PR to switch to the library when things solidify.

Closing this crazy out.

PR #701, opened 2015-10-28
[bug 1216998] Switch to Puente

willkg:
This is a work in progress to use Puente instead of Fjord's l10n extract/merge stuff.


willkg:
I just rebased against master. If this works, then I'll land this and push to stage. Then I'll go through and test the feedback forms for important locales. Once everything is good in that regard, I'll push to production.

willkg:
Everything looks good, so I'm self-r+ing and landing this.

PR #702, opened 2015-10-30
Fix msg length

willkg:
If the msg is too long, then it throws a db error. This truncates the
msg value.

### DIFF
--- a/fjord/journal/models.py
+++ b/fjord/journal/models.py
@@ -12,10 +12,34 @@ RECORD_INFO = u'info'
 RECORD_ERROR = u'error'
 
 
+def goofy_truncate(text, length):
+    """This truncates text accounting for utf-8 sequences
+
+    This assumes a limited number of bytes of storage, but that we
+    want to maintain utf-8 sequences. Thus it does this goofy loop
+    to make sure we don't break a sequence, but that the result
+    is correctly truncated.
+
+    :arg text: either unicode or str
+    :arg length: the number of bytes to truncate to
+
+    :returns: utf-8 encoded string
+    """
+    text = unicode(text)
+
+    while len(text.encode('utf-8')) > length:
+        if len(text) > 255:
+            text = text[:255]
+        else:
+            text = text[:-1]
+    return text.encode('utf-8')
+
+
 class RecordManager(models.Manager):
     @classmethod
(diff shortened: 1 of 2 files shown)


willkg:
@mythmon Can you give this a quick look? Is it sane or a terrible idea? As a side note, I'm not sure what this does to utf-8 encoding.

willkg:
Hrm. I guess I could do a goofy loop:

```
msg = unicode(msg)
while len(msg.encode('utf-8')) > 255:
    msg = msg[:-1]
msg = msg.encode('utf-8')
```

fjord/journal/models.py
```diff
@@ -16,6 +16,8 @@ class RecordManager(models.Manager):
     @classmethod
     def log(cls, type_, app, src, action, msg, instance=None, metadata=None):
         msg = msg.encode('utf-8')
+        if len(msg) > 255:
+            msg = msg[:255]
```
robhudson: If you're worried about unicode you could maybe use the django utils `truncate`? It probably handles it smartly.

willkg:
Yeah, django has a Truncator class which does some clever stuff, but it's locale-aware and does a bunch of other stuff. I'm not wildly excited about using it.

I'll go with the goofy loop. The output is something that only I see. If it's too goofy, I'll fix it later.

willkg:
^^^ That redoes the code to use a loop with tests.

robhudson:
Seems nicely smarter. r+

mythmon:
I found [this Stack Overflow post](http://stackoverflow.com/questions/1809531/truncating-unicode-so-it-fits-a-maximum-size-when-encoded-for-wire-transfer) that suggests another way. `msg.encode('utf-8')[:255].decode('utf-8', 'ignore')`.

That ignores parsing errors and just throws away bad data. In my tests it seems to behave the same as your test case. It might hide other errors though, but as long as `msg` was valid UTF8 before, I think it would be safe.

mythmon:
That being said, the current loop likes it will work fine.

willkg:
I'm going to leave the goofy loop because i need to land and push this out with enough time for me to fix issues because next week is release week and today is friday and OMG THE STRESS!!! THE ANXIETY!!!

But I like your way a lot. I wish I thought of looking at stackoverflow.

willkg:
Thank you both!

PR #703, opened 2015-11-04
[bug 1218381] Fix run_l10n_completion runs

willkg:
It looks like the script is running just fine, but it's putting the
file in the source tree rather than the web tree and that's not
correct. We need to put it in the web tree so that it's available
everywhere.

Pretty sure this should fix it.

r?

### DIFF
--- a/bin/crontab/crontab.tpl
+++ b/bin/crontab/crontab.tpl
@@ -9,7 +9,7 @@ HOME=/tmp
 # Once a day at 2:00am run the l10n completion script.
 # Note: This runs as root so it has access to the locale/ directory
 # to do an svn up.
-10 2 * * * root cd {{ source }} && (./bin/run_l10n_completion.sh {{ source }} {{ python }})
+10 2 * * * root cd {{ source }} && (./bin/run_l10n_completion.sh {{ webapp }} {{ python }})
 
 # Once a day at 3:00am run the translation daily activites.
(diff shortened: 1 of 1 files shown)


willkg:
@mythmon Does this look right? It's curious since as it stands now, I'm not sure how it ever worked, but it definitely did since I used it to track things. Further, the call on kitsune is totally different:

```
30 03 * * * root {{ rscripts }} scripts/l10n_completion.py --truncate 30 locale media/uploads/l10n_history.json media/uploads/l10n_summary.json
```

That puzzles me greatly since I'm pretty sure I wrote both of them.

mythmon:
The new version doesn't look worse than the old one, and that path is indeed the right one. I'm confused by the parenthesis in the cronline, but that's probably not important.

[`bin/run_l10n_completion.sh`](https://github.com/willkg/fjord/blob/49ff6e261a6bb645660b79ae23df435f7a698bde/bin/run_l10n_completion.sh) does a lot of stuff, but one of the things it does looks a lot like Sumo's version of this cron job.

Without looking at git history, I'd imagine that `run_l10n_completion.sh` got changed, and the cronjob didn't get updated. That would explain how it used to work, but doesn't now.

mythmon:
r+

willkg:
I'm suspecting it's a change on the servers. Maybe at one point source == webapp. Hard to say for sure, though. :(

Thank you for looking at it!

PR #704, opened 2015-11-12
[bug 1221252] Fix url field handling for feedback

willkg:
This changes the url field to be 1000 characters. Further, it changes
the API and the feedback form handler to allow urls bigger than 1000
characters, but truncate them. It adds a test for the feedback form
handler and fixes the api test. On top of that, it centralizes the
magic number to the config file so I can more easily change it.

### DIFF
--- a/fjord/feedback/config.py
+++ b/fjord/feedback/config.py
@@ -1,6 +1,10 @@
 from django.utils.translation import ugettext_lazy as _lazy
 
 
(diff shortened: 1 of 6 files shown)


willkg:
This is pretty straight-forward and the changes look decent. I'll wait for Travis to give it the greenlight and then self-r+ it and land it.

willkg:
Green light!

PR #706, opened 2015-11-12
1001430 db strings

willkg:
So many things wrapped in this. Ugh.


willkg:
I'm going to wait for Travis to greenlight this. Then I'll push to stage and verify translations and locale-switching work. Then I'll push to prod.

willkg:
Green!

PR #707, opened 2015-11-16
1221352 desc

willkg:
This adds handling for emoji characters in feedback descriptions.

### DIFF
--- a/fjord/feedback/utils.py
+++ b/fjord/feedback/utils.py
@@ -1,3 +1,10 @@
+import re
+# unicodedata2 is the unicodedata backport to Python 2
+try:
+    import unicodedata2 as unicodedata
+except ImportError:
+    import unicodedata
+
 import urlparse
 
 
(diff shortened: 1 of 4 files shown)


willkg:
@mythmon If you have time today to give this a skim, that'd be super. If not, I'll wait for travis to greenlight it and then push it. I need it fixed today.

fjord/feedback/utils.py
```diff
@@ -28,3 +35,23 @@ def clean_url(url):
...
+
+    """
+    # convert to unicodedata.name()
+    def _convert(match_char):
+        c = match_char.group(0)
+        try:
+            return unicodedata.name(c)
+        except ValueError:
+            # Throws a ValueError if the name doesn't exist.
+            return ''
```
willkg: I thought about having this return "CRYING BECAUSE MYSQL SUCKS" because that's how I feel, but I didn't want to mess up a lot of feedback data and everything sucks--it's not just mysql.

I figured it was prudent to immortalize that in at least a pr comment.

willkg:
So, there's a lot of shenanigans here:

1. mysql doesn't like emoji.
2. unicodedata can convert that to names, but python2's unicodedata doesn't understand, so we have to use unicodedata2 which is a backport from python3.
3. unicodedata2 is a binary dependency which means we need to install it on all the machines we need to use it on and because Input is still using vendor/ and vendor/ is only pure-python dependencies, this means we need to open up a webops bug.
4. the servers are currently frozen, so we can't get unicodedata2 installed until the freeze is lifted.
5. :cry: <-- (something you can't do on Input right now)

fjord/feedback/utils.py
mythmon: This is more of a product decision, but should this be something like "<unknown unicode 0x12345>"? I don't know how much you are trying to preserve data vs just make it work.

willkg: This is frustrating. I feel really bad about throwing this data away, but putting anything else here possibly changes the meaning of things.

For example:

```
I love firefox ???
```

I thought about spending some time now looking for a "good replacement" then decided I could do that change later.

mythmon: :+1:

willkg:
Travis is green and this seems good, so I'm going to land it and push it out now.

PR #708, opened 2015-11-18
[bug 1223625] Tag old fennec feedback with a source

willkg:
Old fennec had an in-product form that used the old Input POST
params. This change adds a "source" to that so we can see how much
feedback we're getting from old fennec which will inform us as to
when we can take this code out.


willkg:
I'll wait for Travis to greenlight this, then I'll land it.

willkg:
Green!

PR #709, opened 2015-11-18
Add source param to secret histogram api

willkg:
This adds a "source" param to the secret histogram api. We can use this for dashboards that want to track how many feedback we're getting broken down by source.


willkg:
I'll wait for Travis to greenlight this, then land it.

willkg:
Closing this for another PR.

PR #710, opened 2015-11-18
[bug 1223625] Add source and api params to secret histogram api

willkg:
I also redid the api code in ES because previously if a feedback
wasn't saved with the API, then the value would be None which ES
doesn't index and it's harder to query. This changes it so that the
api values are all strings and the None value is now "None".


willkg:
I'm going to wait until Travis is green, then land this. After I land it, I'll need to reindex stage and prod because i made a change to ResponseDocType.

willkg:
Green!

PR #714, opened 2015-11-30
[bug 1226772] Upgrade Django to 1.8.7

willkg:


willkg:
Travis is good. This is a minor django upgrade that changes things that Input doesn't really use. So I landed it.

PR #716, opened 2015-12-01
[bug 1229366] Pin hb data view and improve logging

willkg:
Previously, I think it was possible for Input to stomp on a new data
packet with an old one because the updated_ts check would be using on
stale data because we're in a multidb environment and there's
replication lag and all that.

This pins that view so that it always happens on the master db and thus
always sees the current view of the world.

Further, this adds a bunch of statsd calls so we can measure various
scenarios we don't otherwise have a window into.

### DIFF
--- a/fjord/heartbeat/api_views.py
+++ b/fjord/heartbeat/api_views.py
@@ -49,7 +49,7 @@ class HeartbeatV2API(rest_framework.views.APIView):
         # If it's an empty packet, bail immediately with a nicer message.
         if not post_data:
             statsd.incr('heartbeat.emptypacket')
-            return self.rest_error(post_data, ['empty packet'])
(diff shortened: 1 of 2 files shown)


willkg:
There are a couple of problems here. One is that the tests don't cover this change in any useful way. The second is that the theory is that there's a pinning issue, but I don't really have a good way of verifying whether this code fixes the problem or not. I think I need to push it to production and test it out there. The statds logging will tell us some things, but we don't know what life was like before these changes so we can't determine whether this improves or worsens the situation.

Anyhow, yay!

Travis is green so I'm landing it.

willkg:
^^^ Added a test for the "packet is empty" scenario. I'll land it after Travis is good with that.

willkg:
Green--landing!

PR #720, opened 2015-12-14
DO NOT MERGE!!!!!111!!! [bug 1229511] Upgrade to mysql 1.2.5, switch to utf8mb4

willkg:
This upgrades the mysql lib to 1.2.5 which supports the utf8mb4
charset that supports 4-byte utf-8 characters.

When doing that, I was able to remove all the "remove emoji" code.

Further, upgrading the mysql lib caused some tests to break due to
improved precision of datetimes (down to the millisecond now). This
fixes those tests, too.


willkg:
DO NOT MERGE!!!11!!

We can't merge this until mysql python lib has been upgraded on all the servers. That's covered in bug 1232406.

willkg:
Travis is failing because the results don't have milliseconds, but the expected results do. I'm not sure why that would be. When upgrading mysql python lib locally, I suddenly started getting milliseconds.

I'll have to look into that.

willkg:
I did some poking around and I think the problem is that travis runs tests in an Ubuntu 12.04 container which comes with mysql 5.5.46-0ubuntu0.12.04.2. The utf8mb4 stuff shows up in mysql 5.5.3, so that's good. However, mysql 5.6 introduced millisecond support in datetime stamp columns and that's where the tests are failing.

I'll have to do some soul searching as to how I want to deal with that.

willkg:
I added a comment to bug #1221352 which covers the current status of all the work involved including this PR.

The travis issue needs to be fixed before this can go further. A possible option is this: https://github.com/mozilla/treeherder/pull/1080

willkg:
While this solves a bunch of problems, it creates a bunch more. I talked with @Osmose about it and we think we're going to table this for now.

Given that, I'm closing it out.

PR #721, opened 2015-12-17
Implement mailinglist system

willkg:
The mailing list system lets us associate names with lists of
recipients. A list of recipients is stored in a text field that allows
for arbitrary mailing list sizes, but also comments to make
managing the list easier.

### DIFF
--- a/fjord/mailinglist/admin.py
+++ b/fjord/mailinglist/admin.py
@@ -1,0 +1,10 @@
+from django.contrib import admin
+
+from fjord.mailinglist.models import MailingList
+
+
+class MailingListAdmin(admin.ModelAdmin):
+    list_display = ('name', 'members')
+
+
+admin.site.register(MailingList, MailingListAdmin)
--- a/fjord/mailinglist/models.py
+++ b/fjord/mailinglist/models.py
@@ -1,0 +1,30 @@
+from django.db import models
+
+from fjord.base.models import ModelBase
+
+
+class MailingList(ModelBase):
+    name = models.CharField(
+        max_length=100, unique=True,
+        help_text=u'Unique name to identify the mailing list'
+    )
+    members = models.TextField(
+        blank=True, default=u'',
+        help_text=(
+            u'List of email addresses for the list--one per line. You '
+            u'can do blank lines. # and everything after that character '
+            u'on the line denotes a comment.'
+        )
+    )
+
+    @property
+    def recipient_list(self):
+        members = []
+        for line in self.members.splitlines():
+            if '#' in line:
+                line = line[:line.find('#')]
+            line = line.strip()
+            if line:
+                members.append(line)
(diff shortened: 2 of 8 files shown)


willkg:
@mythmon @Osmose This is a very light-weight mailing list thing. It's not used, yet, but I'm going to switch the gengo account balance notification system to use it and I'm going to use it for the heartbeat health check emails.

Does this look "good enough" for now?

mythmon:
> Every program attempts to expand until it can read mail. Those programs which cannot so expand are replaced by ones which can.
> -[Jamie Zawinski](https://www.jwz.org/hacks/)

willkg:
@mythmon said he'd look at this some time today. I'll land it tomorrow regardless. I'm pretty sure it's simple and not wildly exciting. But I'm also not going to maintain it, so it's worth getting external input.

willkg:
I talked with @mythmon and he's swamped. There's one model in this and I'm planning to use it, but not heavily and the API points are really clear, so even if this is a terrible thing, it'll be easy to fix in the future.

Given that and Travis' lovely green evening gown, I'm going to land this.

fjord/mailinglist/admin.py
```diff
@@ -0,0 +1,10 @@
+from django.contrib import admin
+
+from fjord.mailinglist.models import MailingList
+
+
+class MailingListAdmin(admin.ModelAdmin):
+    list_display = ('name', 'members')
+
+
+admin.site.register(MailingList, MailingListAdmin)
```
Osmose: There's also [the decorator](https://docs.djangoproject.com/en/1.8/ref/contrib/admin/#the-register-decorator) if you're into that.

fjord/mailinglist/models.py
```diff
@@ -0,0 +1,30 @@
...
+
+class MailingList(ModelBase):
+    name = models.CharField(
+        max_length=100, unique=True,
+        help_text=u'Unique name to identify the mailing list'
+    )
+    members = models.TextField(
+        blank=True, default=u'',
+        help_text=(
+            u'List of email addresses for the list--one per line. You '
```
Osmose: Someday someone will write a version of `dedent` that also joins the lines together. And on that day, we will rejoice.

But not today!

Osmose:
Out of context, this looks fine to me! Nice work!

fjord/mailinglist/admin.py
willkg: I don't know what I've been doing wrong, but I can't get the decorator to work. I use it, but it doesn't show up in the admin. I never spent the time to look into what was going on, though.

Osmose: You might have to explicitly pass the admin site instance if you're monkeypatching the admin site with a new one, depending on the order of imports and monkeypatching.

PR #722, opened 2015-12-18
[bug 1225676] {WIP don't merge} Complete backend steel thread

hoosteeno:
I made changes to the view here that I think will enable custom android input and thanks templates without breaking things. Feedback welcome!


fjord/feedback/views.py
```diff
@@ -135,10 +142,10 @@ def _handle_feedback_post(request, locale=None, product=None,
     data = form.cleaned_data
 
     description = data.get('description', u'').strip()
+
     if not description:
-        # If there's no description, then there's nothing to do here,
-        # so thank the user and move on.
-        return HttpResponseRedirect(reverse('thanks'))
+        # If there's no description populate a generic one
+        description = u'Sentiment: %s' % (('Sad', 'Happy')[data['happy']])
```
willkg: Input feedback isn't a rating system--it's a system for gathering textual sentiment which requires descriptions. That's why if there's no description, we just drop it and move on.

Do you have a requirement for feedback with no descriptions? If so, that'll require other changes including to the ratelimiting which will likely have some big ramifications.

fjord/feedback/views.py
```diff
@@ -470,8 +498,7 @@ def feedback_router(request, product=None, version=None, channel=None,
     elif product in PRODUCT_OVERRIDE:
         # If the product is really a form name, we use that
         # form specifically.
-        view = PRODUCT_OVERRIDE[product]
-        product = None
+        view = PRODUCT_OVERRIDE[product]['view']
```
willkg: I think you're not using this right. The PRODUCT_OVERRIDE is really an (ab)use of the product field in the url to direct to a specific form. In that case, the product variable holds a form name--not a product.

I think instead of doing this, you should follow the firefoxos path in the previous `if` block where it checks and does something different that's product-specific.

fjord/feedback/views.py
```diff
@@ -72,11 +72,18 @@ def thanks(request):
...
         suggestions = get_suggestions(feedback, request)
 
-    return render(request, 'feedback/thanks.html', {
-        'product': product,
+    product = models.Product.objects.get(db_name=product_db_name)
+
+    template = 'feedback/thanks.html'
+
+    if product.slug in PRODUCT_OVERRIDE:
+      template = PRODUCT_OVERRIDE[product.slug]['thanks_template']
```
willkg: Nit: We use four spaces--not two.

Also, instead of doing this here, you should pass the thank you page in via the args.

willkg: Scratch that--I misread the context. My bad. Mmm... Is the android thank you page different than the regular one? Why?

hoosteeno: The fennec input and thank you pages have their own style (see screenshots here: https://docs.google.com/document/d/1yROux6xW2vzjpV9OYx9ppOtdYm-Mu3rApjZ3C-8xGLg/edit#bookmark=id.k0hw2dxlnutc) and content; they don't take Input's standard style or content.

And this project's ultimate intent is to make optimizing them easier, which requires that they be isolated from the generic templates.

Regarding two/four spaces, that's my bad. My editor doesn't switch indent styles automatically. I will definitely run code through flake8 before final PR, so you can probably ignore nits for now. :)

fjord/feedback/views.py
hoosteeno: I don't love this approach, but it did seem like the least disruptive. The screenshots hopefully explain why we need it: When Fennec users say "It made me happy," they don't get an input form. A couple thoughts on this:
1) I don't think we would want that input form to be required, even if we did include it. It's simply not as likely to attract description content due to thumbtyping. 
2) I don't think we should include it. Our hope is to send happy mobile users to the application store where they can give us 5 stars and create a growth loop. 

Considering those, a couple ideas. Which do you like better?

a) Drop "happy" mobile feedback on the floor. Use the form only as a way route mobile users to the app store.
b) Help me dig into the bigger changes that would allow us to capture happy mobile feedback without requiring description content.

I think the answer probably depends on Input's overarching reason to exist. Is it to capture all product feedback, or just qualitative feedback?

fjord/feedback/views.py
willkg: None of those screenshots show the thank you page. I thought you were just going to use the Input thank you page as is. It'd be a lot easier to do that.

I'm not sure I'd (ab)use the `PRODUCT_OVERRIDE` thing for this, though. It'd be better to fix the way the pieces connect together than do workarounds like this.

Regarding style issues, we have an editorconfig thing. If you can configure your editor to use that, it sometimes makes it easier. http://editorconfig.org/

fjord/feedback/views.py
willkg: Input gathers feedback for sentiment analysis which requires text content. In order to accept feedback without descriptions, we'd need to do a major overhaul of Input or create a parallel feedback system. That's a big project and has a lot of ramifications.

Thinking about the options a bit, we can't send them to the apple store *and* also have them leave feedback in Input. We also can't forward feedback from Input to the apple store. Given that we want happy feedback in the apple store, I think the only option here is if the user says "happy", we give them an interstitial page that links to the apple store and that's the end of the flow. The form shouldn't POST anything to Input. The user never sees the "thank you" page.

hoosteeno:
Closing; new pull request addressing these comments forthcoming.

PR #723, opened 2015-12-24
[bug 1225676] Port fennec feedback form to web

hoosteeno:
This PR does 2 things. 
1. Includes front-end integration of styles from about:feedback in FxA. Still missing links and integration with "later" and "never". Check the SVG hotness.
2. Addresses the concerns from #722. Specifically...

* doesn't overload PRODUCT_OVERRIDE
* doesn't customize /thanks
* doesn't save happy feedback

### DIFF
--- a/fjord/feedback/jinja2/feedback/android_feedback.html
+++ b/fjord/feedback/jinja2/feedback/android_feedback.html
@@ -197,8 +197,8 @@
           </g>
         </g>
       </svg>
-      {% trans %}
-        Visit our <a href="https://support.mozilla.org/products/mobile/popular-articles-android?as=u&utm_source=inproduct">Support Site</a> to get answers for common issues.
+      {% trans support_url="https://support.mozilla.org/products/mobile/popular-articles-android?as=u&utm_source=inproduct" %}
+        Visit our <a href="{{support_url}}">Support Site</a> to get answers for common issues.
         {% endtrans %}
     </div>{# .help-section #}
   </div>{# .footer.sad #}
--- a/fjord/feedback/jinja2/feedback/android_feedback.html
+++ b/fjord/feedback/jinja2/feedback/android_feedback.html
@@ -1,0 +1,205 @@
+{% extends "base.html" %}
+
+{% block extra_headers %}
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no" />
+{% endblock %}
+
+{% block site_js %}
+  {% javascript 'android_feedback' %}
+{% endblock %}
+
+{% block site_css %}
+  {% stylesheet 'android-feedback' %}
+{% endblock %}
+
+{# L10N note: In order to reduce the work required by translators
+ # to launch the initial port of this form, we copied localized
+ # strings directly from Fennec (see bug 1235419). We can/should
+ # fix strings after launch.
+ #}
+{% block page_title %}{{ _('Firefox Feedback') }}{% endblock %}
+
+{% set extra_body_attrs =
+  {
+    'data-form-name': 'android',
+    'data-version': version,
+    'data-channel': channel
+  }
+%}
+
+{% block body_id %}android{% endblock %}
+
+{% block body %}
+  <div class="main">
+    {% block content %}
+        <div class="content">
+
+          <div class="deck">
+            <div class="card no-back" id="intro">
+              <section>
+
+                <h1>
+                  {% trans %}
+                    Have a minute?
+                  {% endtrans %}
+                </h1>
+                <div class="message">
+                  {% trans product=product.display_name %}
+                    Tell us what you think about {{product}} so far.
+                  {% endtrans %}
+                </div>
+
+                <noscript>
+                  <p class="error">
+                    {% trans %}
+                      JavaScript is required to leave feedback. Please enable JavaScript in your browser and refresh this page.
+                    {% endtrans %}
+                  </p>
+                </noscript>
+
+                <div id="happy-button" class="link-box sentiment-button" tabindex=0>
+                  <a>
+                    {% trans %}
+                      I love it
+                    {% endtrans %}
+                  </a>
+                </div>{# #happy-button #}
+
+                <div id="sad-button" class="link-box sentiment-button" tabindex=0>
+                  <a>
+                    {% trans %}
+                      I ran into some problems
+                    {% endtrans %}
+                  </a>
+                </div>{# #sad-button #}
+
+              </section>
+            </div>{# #intro.card #}
+
+            <div class="card inactive" id="moreinfo" data-focus="#description" data-back-id="intro">
+              <section>
+                <div class="happy">
+                  <h1>
+                    {% trans %}
+                      That's great to hear!
+                    {% endtrans %}
+                  </h1>
+
+                  <div class="message-box">
+                    <div class="message">
+                      {% trans %}
+                        Want to share the love by giving us a 5 star rating on
+                        Google Play?
+                      {% endtrans %}
+                    </div>{# .message #}
+                    <div>
+                      {% trans %}
+                        It takes less than a minute and feels great.
+                      {% endtrans %}
+                    </div>
+                    <a class="stars" id="play-store" data-event="FeedbackOpenPlay" href="#">
+                      <svg>
+                        <defs>
+                          <symbol viewBox="0 0 100 100" id="star">
+                            <polygon points="48 75 18.6107374 90.4508497 24.2235871 57.7254249 0.447174185 34.5491503 33.3053687 29.7745751 48 0 62.6946313 29.7745751 95.5528258 34.5491503 71.7764129 57.7254249 77.3892626 90.4508497" class="star" xmlns="http://www.w3.org/2000/svg" fill="#0092db"></polygon>
+                          </symbol>
+                        </defs>
+                        <use width="25" height="25" xlink:href="#star" x="0" y="0"></use>
+                        <use width="25" height="25" xlink:href="#star" x="25" y="0"></use>
+                        <use width="25" height="25" xlink:href="#star" x="50" y="0"></use>
+                        <use width="25" height="25" xlink:href="#star" x="75" y="0"></use>
+                        <use width="25" height="25" xlink:href="#star" x="100" y="0"></use>
+                      </svg>
+                      {% trans %}
+                        Yes, go to Google Play
+                      {% endtrans %}
+                    </a>
+                  </div>{# .message-box #}
+                </div>{# .happy #}
+
+                <div class="sad">
+                  <h1>
+                    {% trans %}
+                      Oh no!
+                    {% endtrans %}
+                  </h1>
+
+                  <div class="ask">
+                    <form id="responseform" action="" method="post">
+                      <div class="input-box">
+                        <label for="description">
+                          {% trans %}
+                            We're sorry that you had some problems with Firefox.
+                            Please tell us what happened so that we can fix it.
+                          {% endtrans %}
+                        </label>
+                        <div id="description-counter">{{ TRUNCATE_LENGTH }}</div>
+                        <textarea data-max-length="{{ TRUNCATE_LENGTH }}" id="description" name="description" placeholder="Enter your feedback here"></textarea>
+                      </div>{# .input-box #}
+
+                      <div class="input-box">
+                        <label for="id_url">
+                          {% trans %}
+                            Include last visited site
+                          {% endtrans %}
+                          <input id="last-checkbox" type="checkbox" checked="checked"/>
+                          <input class="url" id="id_url" name="url" placeholder="http://" type="text">
+                        </label>
+                      </div>{# .input-box #}
+
+                      {% for hidden in form.hidden_fields() %}
+                        {{ hidden }}
+                      {% endfor %}
+                      {% csrf_token %}
+
+                      <div class="input-box">
+                        <input type="submit" value="{{ _('Send Feedback') }}" id="form-submit-btn" class="complete btn submit" disabled="disabled">
+                        </input>
+                      </div>{# .input-box #}
+
+                      <div class="privacy">
+                        {% trans %}
+                          For your privacy, please do not include any personal information
+                          in your feedback.
+                        {% endtrans %}
+                      </div>{# .privacy #}
+                    </form>
+                  </div>{# .ask #}
+
+                </div>{# .sad #}
+              </section>
+            </div>{# #moreinfo.card #}
+          </div>{# .deck #}
+        </div>{# .content #}
+
+    {% endblock %}
+  </div>
+  <div class="footer happy">
+    <a id="maybe-later" data-event="FeedbackMaybeLater" href="#">
+      {% trans %}
+        Maybe later
+      {% endtrans %}
+    </a>
+    <a id="no-thanks" data-event="FeedbackClose" href="#">
+      {% trans %}
+        No thanks
+      {% endtrans %}
+    </a>
+  </div>{# .footer.happy #}
+  <div class="footer sad">
+    <div class="help-section">
+      <svg viewBox="0 0 156 156">
+        <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+          <g transform="translate(3.000000, 2.000000)">
+              <ellipse stroke="#DE1920" stroke-width="6" fill="#FFFFFF" transform="translate(75.008751, 75.991249) rotate(-270.000000) translate(-75.008751, -75.991249) " cx="75.008751" cy="75.991249" rx="75.008751" ry="75.008751"></ellipse>
+              <path d="M114.879498,75.7124061 L150.016995,75.7124061 C149.8667,34.4144668 116.341953,0.982498052 75.008751,0.982498052 C74.7295491,0.982498052 74.4507036,0.98402351 74.1722222,0.987066548 L74.1722222,36.1205023 C74.357816,36.1179759 74.543714,36.1167085 74.729908,36.1167085 C96.7198153,36.1167085 114.581115,53.7933778 114.879498,75.7124061 Z M74.729908,116.423475 L74.729908,150.999529 C33.4319688,150.849198 0,117.324451 0,75.991249 C0,75.8982618 0.000169204083,75.805314 0.000507321153,75.7124061 L34.5803184,75.7124061 C34.5777919,75.8979999 34.5765246,76.083898 34.5765246,76.270092 C34.5765246,98.4461933 52.5538067,116.423475 74.729908,116.423475 Z" fill="#DE1920" transform="translate(75.008497, 75.990995) rotate(-270.000000) translate(-75.008497, -75.990995) "></path>
+              <ellipse stroke="#DE1920" stroke-width="6" fill-opacity="0" transform="translate(74.729908, 76.270092) rotate(-270.000000) translate(-74.729908, -76.270092) " cx="74.729908" cy="76.270092" rx="40.1533834" ry="40.1533834"></ellipse>
+          </g>
+        </g>
+      </svg>
+      {% trans %}
+        Visit our <a href="https://support.mozilla.org/products/mobile/popular-articles-android?as=u&utm_source=inproduct">Support Site</a> to get answers for common issues.
+        {% endtrans %}
+    </div>{# .help-section #}
+  </div>{# .footer.sad #}
+{% endblock %}
--- a/fjord/feedback/jinja2/feedback/android_thanks.html
+++ b/fjord/feedback/jinja2/feedback/android_thanks.html
@@ -1,0 +1,76 @@
+{% extends "base.html" %}
+
+{% block extra_headers %}
+  <meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no" />
+{% endblock %}
+
+{% block site_js %}
+  {% javascript 'android_feedback' %}
+{% endblock %}
+
+{% block site_css %}
+  {% stylesheet 'android-feedback' %}
+{% endblock %}
+
+{# L10N note: In order to reduce the work required by translators
+ # to launch the initial port of this form, we copied localized
+ # strings directly from Fennec (see bug 1235419). We can/should
+ # fix strings after launch.
+ #}
+{% block page_title %}{{ _('Firefox Feedback') }}{% endblock %}
+
+{% block body_id %}android{% endblock %}
+
+{% block body %}
+  <div class="main">
+
+    {% block content %}
+      <div class="content">
+        <div class="deck">
+          <div class="card no-back" id="thanks">
+            <section>
+
+              <h1>
+                {% trans %}
+                  Thank you.
+                {% endtrans %}
+              </h1>
+              <div class="message heart">
+                <svg viewBox="0 0 100 110">
+                  <path d="M4.20152102,44.5344113 C-12.8987227,5.08506476 37.2648073,-17.6397688 50.6373929,18.0189693 C63.9576781,-17.6397683 114.277469,5.08506476 97.1772249,44.5344113 C91.8645859,55.9698934 52.2858248,83.7136852 50.5850925,90.6483352 C48.9013298,83.7202158 9.80833953,55.370549 4.20152102,44.5344113 Z" fill="#DE1920"></path>
+                </svg>
+                {% trans %}
+                  We're always working to make Firefox better.
+                  Rest assured that real people will look at your feedback 
+                  and do their very best to resolve your issue.
+                {% endtrans %}
+              </div>
+              <div class="message">
+                {% trans %}
+                  Or else.
+                {% endtrans %}
+              </div>
+            </section>
+          </div>{# #thanks #}
+        </div>{# .deck #}
+      </div>{# .content #}
+    {% endblock %}
+
+  </div>{# .main #}
+  <div class="footer sad show">
+    <div class="help-section">
+      <svg viewBox="0 0 156 156">
+        <g stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
+          <g transform="translate(3.000000, 2.000000)">
+              <ellipse stroke="#DE1920" stroke-width="6" fill="#FFFFFF" transform="translate(75.008751, 75.991249) rotate(-270.000000) translate(-75.008751, -75.991249) " cx="75.008751" cy="75.991249" rx="75.008751" ry="75.008751"></ellipse>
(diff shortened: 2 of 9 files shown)


willkg:
To summarize what I said on IRC, I think this touches a bunch of crappy code to deal with flows through leaving feedback. I'm going to write up a bug to fix that. I'm pretty sure I can fix it in a way that's more loosely coupled, easier to follow and thus easier to maintain going forward. I'll work on that today.

willkg:
Plus, I'll run any changes by @hoosteeno before I commit anything.

willkg:
I refactored and cleaned up a bunch of feedback code. That should reduce the changes this PR needs to do in `fjord/feedback/views.py` and make it easier/possible to change the Thank You page template--something @hoosteeno had in a previous PR.

@hoosteeno: Feel free to ditch this PR and start a new one or force-push on this one--whatever is easiest for you.

fjord/feedback/jinja2/feedback/android_feedback.html
```diff
@@ -0,0 +1,192 @@
...
+
+{% block site_js %}
+  {% javascript 'generic_feedback' %}
+{% endblock %}
+
+{% block site_css %}
+  {% stylesheet 'android-feedback' %}
+{% endblock %}
+
+{% block page_title %}{{ _('Firefox for Android Feedback') }}{% endblock %}
```
willkg: We should add Jinja2 comment before this line explaining the l10n situation so that future developers understand what's going on. Namely that since we're copying strings over from Firefox for Android's codebase, we need to use the exact same strings.

fjord/feedback/jinja2/feedback/android_feedback.html
```diff
@@ -0,0 +1,192 @@
...
+                        </defs>
+                        <use width="25" height="25" xlink:href="#star" x="0" y="0"></use>
+                        <use width="25" height="25" xlink:href="#star" x="25" y="0"></use>
+                        <use width="25" height="25" xlink:href="#star" x="50" y="0"></use>
+                        <use width="25" height="25" xlink:href="#star" x="75" y="0"></use>
+                        <use width="25" height="25" xlink:href="#star" x="100" y="0"></use>
+                      </svg>
+                      {% trans %}
+                        Yes, go to Google Play
+                      {% endtrans %}
```
willkg: Will the above work for non-Firefox for Android browsers? What happens when someone using Firefox desktop looks at this form?

fjord/feedback/jinja2/feedback/android_feedback.html
hoosteeno: Yeah, can do. And looking at it again I see the title is actually "Firefox Feedback", so I'll adjust.

fjord/feedback/views.py
```diff
@@ -98,7 +98,7 @@ def thanks_view(request):
             pass
 
     if feedback:
-        product_slug = feedback.product
```
hoosteeno: All the remediations for slug/name should be backed out until code is rebased onto code from bug 1236581.

fjord/feedback/jinja2/feedback/android_thanks.html
```diff
@@ -0,0 +1,71 @@
...
+              </h1>
+              <div class="message heart">
+                <svg viewBox="0 0 100 110">
+                  <path d="M4.20152102,44.5344113 C-12.8987227,5.08506476 37.2648073,-17.6397688 50.6373929,18.0189693 C63.9576781,-17.6397683 114.277469,5.08506476 97.1772249,44.5344113 C91.8645859,55.9698934 52.2858248,83.7136852 50.5850925,90.6483352 C48.9013298,83.7202158 9.80833953,55.370549 4.20152102,44.5344113 Z" fill="#DE1920"></path>
+                </svg>
+                {% trans %}
+                  We're always working to make Firefox better.
+                  Rest assured that real people will look at your feedback 
+                  and do their very best to resolve your issue.
+                {% endtrans %}
```
willkg: This means we're not showing Thank You suggestions which the view code is generating. I suspect they'd be helpful.

I think with the thank you page, we care less about whether things are translated than we do with the feedback form. I claim we're ok if we have new strings here. If that's not true, then we should have a follow-up bug for redoing the thank you page to add the suggestions section.

fjord/feedback/jinja2/feedback/android_feedback.html
hoosteeno: Javascript now includes logic to switch this link based on a url parameter we're getting from the in-product feedback page: https://github.com/mozilla/fjord/pull/723/files#diff-aacf79c42cfad2e29069461b3eb426a8R9

If parameter is present, URL is an app intent; otherwise it's http.

fjord/feedback/jinja2/feedback/android_thanks.html
hoosteeno: Do you mean the suggestions to go to the support site, follow us on Twitter, or download a different version of Firefox? I think those are all great candidates for optimization, and I hope :adavis will consider adding them. For now I think our original plan, to exactly replicate the existing input experience, makes sense. But I've filed a follow-up bug:

https://bugzilla.mozilla.org/show_bug.cgi?id=1238775

willkg: Those actually aren't suggestions. Those are links that appear on the generic thank you page regardless of what the feedback is. The suggestions framework looks at the actual feedback and generates a set of suggested links based on the feedback that we think will help the user or help us. These include suggestions of KB articles from SUMO based on the feedback content and trigger-keyword based suggestions.

Here's a screenshot:

http://bluesock.org/~willkg/blog/images/input_thank_you_suggestions.png

fjord/feedback/jinja2/feedback/android_feedback.html
```diff
@@ -0,0 +1,192 @@
...
+
+{% block site_js %}
+  {% javascript 'android_feedback' %}
+{% endblock %}
+
+{% block site_css %}
+  {% stylesheet 'android-feedback' %}
+{% endblock %}
+
+{% block page_title %}{{ _('Firefox Feedback') }}{% endblock %}
```
willkg: You should throw a Jinja2 comment before this line explaining the l10n situation in this template. Something along these lines:

```
{#
 # Note on L10N: In order to reduce the work required by translators for us to get this 
 # first phase done, we've copied the strings from Firefox for Fennec exactly. We should
 # fix the strings at some point after this is in production. See bug #xxxxxxx.
 #}

File: fjord/feedback/jinja2/feedback/android_thanks.html (hunk 4 above)
Comment (hoosteeno) [inline fjord/feedback/jinja2/feedback/android_thanks.html]: Oh, cool. I have never seen those before -- there must not be enough SUMO articles about "foo" and "bar". Will update the bug.

File: fjord/feedback/jinja2/feedback/android_feedback.html [hunk 6]
```diff
@@ -0,0 +1,192 @@
...
+                  </p>
+                </noscript>
+
+                <div id="happy-button" class="link-box sentiment-button" tabindex=0>
+                  <a>
+                    {% trans %}
+                      I love it
+                    {% endtrans %}
+                  </a>
+                </div><!-- #happy-button -->
```
willkg: Instead of making these HTML comments which will show up in the output, you could make them Jinja2 comments which won't.

fjord/feedback/jinja2/feedback/android_feedback.html
```diff
@@ -0,0 +1,192 @@
...
+                      </div><!-- .input-box -->
+
+                      <label for="id_url">
+                        {% trans %}
+                          If your feedback is related to a website, you can include it here:
+                        {% endtrans %}
+                      </label>
+                      <div class="input-box">
+                        <input class="url" id="id_url" name="url" placeholder="http://" type="text">
+                      </div><!-- .input-box -->
```
willkg: I'm pretty sure this is different than the existing feedback form in Fennec. In Fennec, there's a checkbox that says "Include last visited site" and lists the last visited site.

We'd need Fennec to pass that along somehow because only they know what it would be.

fjord/feedback/jinja2/feedback/android_feedback.html
```diff
@@ -0,0 +1,192 @@
...
+              <ellipse stroke="#DE1920" stroke-width="6" fill-opacity="0" transform="translate(74.729908, 76.270092) rotate(-270.000000) translate(-74.729908, -76.270092) " cx="74.729908" cy="76.270092" rx="40.1533834" ry="40.1533834"></ellipse>
+          </g>
+        </g>
+      </svg>
+      {% trans %}
+        Visit our <a href="https://support.mozilla.org/en-US/products/mobile/popular-articles-android?as=u&utm_source=inproduct">Support Site</a> to get answers for common issues.
```
willkg: You shouldn't hard-code the locale here. This SUMO link should be:

https://support.mozilla.org/products/mobile/popular-articles-android?as=u&utm_source=inproduct

fjord/feedback/jinja2/feedback/android_thanks.html
```diff
@@ -0,0 +1,71 @@
...
+
+{% block site_js %}
+  {% javascript 'android_feedback' %}
+{% endblock %}
+
+{% block site_css %}
+  {% stylesheet 'android-feedback' %}
+{% endblock %}
+
+{% block page_title %}{{ _('Firefox for Android Feedback') }}{% endblock %}
```
willkg: Should the title of this match the title of the feedback form?

Also, we should include the same L10N message here.

willkg:
I haven't tested this out. Before this lands, we'll want to write up Django tests and smoketests for it.

The Django tests for feedback views are in `fjord/feedback/tests/test_views.py`. Those should be straight-forward, but if you need help writing them, let us know.

The smoketest are in `smoketests/tests/`. Those are a little harder to write. Further, they're only going to test the scenario where a user is using Firefox desktop to fill out the Firefox for Android form. I don't know offhand how to automatically test the scenario where someone is filling it out with Fennec or the scenario where Fennec prompts the user to fill out feedback. We'll need to figure that out. Maybe it's ok to do manual testing for now and then figure out a better solution soon.

Looks good as far as I can tell!

hoosteeno:
Great!
Simulating different devices is as easy as changing URL params. That's not the most robust way to detect devices, but I think it is probably adequate for launch. If you agree, testing will be straightforward.

willkg:
If you can do it solely with urls, that'd be awsome! +1 for that.

fjord/feedback/jinja2/feedback/android_thanks.html
willkg: Ha! That'll do it!

I usually test feedback with "video never works. the screen is always black." It needs to be > 7 words long and I like to throw in "video" because that brings back a bunch because video is awesome.

fjord/feedback/jinja2/feedback/android_feedback.html
```diff
@@ -0,0 +1,199 @@
...
+                        <textarea data-max-length="{{ TRUNCATE_LENGTH }}" id="description" name="description" placeholder="Enter your feedback here"></textarea>
+                      </div>{# .input-box #}
+
+                      <div class="input-box">
+                        <label for="id_url">
+                          {% trans %}
+                            Include last visited site
+                          {% endtrans %}
+                          <input id="last-checkbox" type="checkbox" checked="checked"/>
+                          <input class="url" id="id_url" name="url" placeholder="http://" type="text" readonly="readonly">
```
hoosteeno: I filed a bug to get this from the url param https://bugzilla.mozilla.org/show_bug.cgi?id=1239513

fjord/feedback/static/js/android_feedback.js
```diff
@@ -0,0 +1,43 @@
+(function($) {
+  'use strict';
+
+  var url = decodeURIComponent(window.location.href);
+  var browserReviewUrl = 'https://play.google.com/store/apps/details?id=org.mozilla.CHANNEL#details-reviews';
+  var deviceReviewUrl = 'market://details?id=org.mozilla.CHANNEL';
+  var reviewUrl = '';
+
+  /* decide whether they're on Fx Android or not */
+  var fromDevice = /source=feedback-prompt/.test(url);
```
hoosteeno: I filed a bug to make sure we have the params this script expects: https://bugzilla.mozilla.org/show_bug.cgi?id=1239513

hoosteeno:
@willkg, at long last I have added smoke and unit tests to this code. It is the first time I've done such a thing, so I advise intense and ruthless scrutiny.

willkg:
I thought I was going to get to this today, but aliens ate my babysitter. I'll definitely get to this this week.

fjord/feedback/jinja2/feedback/android_thanks.html
```diff
@@ -0,0 +1,76 @@
...
+              <ellipse stroke="#DE1920" stroke-width="6" fill-opacity="0" transform="translate(74.729908, 76.270092) rotate(-270.000000) translate(-74.729908, -76.270092) " cx="74.729908" cy="76.270092" rx="40.1533834" ry="40.1533834"></ellipse>
+          </g>
+        </g>
+      </svg>
+      {% trans %}
+        Visit our <a href="https://support.mozilla.org/en-US/products/mobile/popular-articles-android?as=u&utm_source=inproduct">Support Site</a> to get answers for common issues.
```
Osmose: This URL also needs the locale removed.

Actually, if you're removing the locale and thus changing the string, you might as well make it a parameter anyway, which is generally how we do URLs in strings. That way we can alter the URL while retaining the translation in the future. Same goes for the other strings with URLs in them.

fjord/feedback/static/js/android_feedback.js
```diff
@@ -0,0 +1,66 @@
+(function($) {
+  'use strict';
+
+  // settings
+  var browserReviewUrl = 'https://play.google.com/store/apps/details?id=org.mozilla.CHANNEL#details-reviews';
+  var deviceReviewUrl = 'market://details?id=org.mozilla.CHANNEL';
+
+  // initialize
+  var reviewUrl, lastUrl = '';
+  var url = decodeURIComponent(window.location.href);
```
Osmose: Probably wanna use `decodeURI` since you're decoding an entire URL.

fjord/feedback/static/js/android_feedback.js
```diff
@@ -0,0 +1,66 @@
+(function($) {
+  'use strict';
+
+  // settings
+  var browserReviewUrl = 'https://play.google.com/store/apps/details?id=org.mozilla.CHANNEL#details-reviews';
+  var deviceReviewUrl = 'market://details?id=org.mozilla.CHANNEL';
+
+  // initialize
+  var reviewUrl, lastUrl = '';
```
Osmose: Style nit, it's generally preferred to put variable declarations on separate lines with separate vars. It makes it more obvious that, in this case, `reviewUrl` is `undefined` rather than an empty string.

fjord/feedback/static/js/android_feedback.js
```diff
@@ -0,0 +1,66 @@
...
+    (Fx Android's feedback prompt includes a utm param)
+  */
+  var fromDevice = /source=feedback-prompt/.test(url);
+  if (fromDevice) {
+    reviewUrl = deviceReviewUrl;
+    /* 
+      TODO: When bug 1237388 is resolved, remove #play-store here.
+      If url params are as assumed below, it'll Just Work(tm)!
+    */
+    $('#maybe-later, #no-thanks, #play-store').click(
```
Osmose: Using `$().on` is preferred over `click` since it handles the event as it bubbles and can deal with JS changing the page:

```js
$(document).on('click', '#maybe-later, #no-thanks, #play-store', function(e) {

});
```

fjord/feedback/static/js/android_feedback.js
```diff
@@ -0,0 +1,66 @@
...
+      function(e) {
+        e.preventDefault();
+        var feedbackEvent = document.createEvent('Events');
```
Osmose: `createEvent` is deprecated. You'll want to use [the Event constructor](https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Creating_and_triggering_events), although I don't know enough about Android handling these events to know if that will work.

fjord/feedback/static/js/android_feedback.js
```diff
@@ -0,0 +1,66 @@
...
+    $('#maybe-later, #no-thanks').hide();
+  }
+
+  // decide what product they might review on play store
+  var version = $('body').attr('data-version');
+  var channel = $('body').attr('data-channel');
+  if (channel == 'beta') {
+    reviewUrl = reviewUrl.replace('CHANNEL', 'firefox_beta');
+  }
+  else if (channel != '' && /^(?!(beta|release)).*$/.test(channel)) {
```
Osmose: I generally prefer using things like `startsWith` instead of regexes when available, as I think it's clearer to read. But that is also totally a personal preference thing; feel free to leave this if you disagree.

fjord/feedback/static/js/android_feedback.js
```diff
@@ -0,0 +1,66 @@
...
+  else {
+    reviewUrl = reviewUrl.replace('CHANNEL', 'firefox');
+  }
+
+  $('#play-store').attr('href', reviewUrl);
+
+  // if they say don't send url in sad feedback, don't send url
+  $('#last-checkbox').click(function(e) {
+    if (this.checked) {
+      $('#id_url').val(lastUrl).prop('disabled', false);
```
Osmose: You should probably store `$('#id_url')` in a variable somewhere to avoid jQuery re-finding it every time you do something to it.

smoketests/pages/android_feedback_form.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class AndroidFeedbackFormPage(FeedbackPage):
+    _page_title = 'Firefox Feedback :: Firefox Input'
+
+    def go_to_feedback_page(self,
+                            version="",
+                            channel="",
+                            last_visited="",
+                            in_device=False):
```
Osmose: This is a very odd kwarg style, you should probably stick to keeping them on one, maybe two lines.

Also, you're using double quotes for some of the strings here when you should be using single quotes.

smoketests/pages/android_feedback_form.py
```diff
@@ -0,0 +1,153 @@
...
+        url = self.base_url + '/feedback/android'
+        params = []
+
+        if version:
+            url = '%s/%s' % (url, version)
```
Osmose: You can just append to the URL instead of formatting.

smoketests/pages/android_feedback_form.py
```diff
@@ -0,0 +1,153 @@
...
+        if in_device:
+            params.append('utm_source=feedback-prompt')
+
+        if last_visited:
+            params.append('url=%s' % (last_visited))
+
+        if len(params) > 0:
+            url = '%s/?%s' % (url, '&'.join(params))
+
+        print('Requesting url: %s' % url)
```
Osmose: You don't generally want to print during tests. Also `print` as a function isn't available in Python 2.7 without a special import.

smoketests/pages/android_feedback_form.py
```diff
@@ -0,0 +1,153 @@
...
+
+        self.selenium.get(url)
+        self.is_the_current_page
+
+    # Intro card
+    _happy_button_locator = (By.CSS_SELECTOR, '#intro #happy-button')
```
Osmose: The rest of the smoketests group the locators at the top, probably want to match their style.

Osmose: Also you can use the `By.ID` selector instead and just use the `happy-button` name since it's an ID.

smoketests/pages/android_feedback_form.py
```diff
@@ -0,0 +1,153 @@
...
+
+    # Happy thanks
+    _playstore_locator = (By.CSS_SELECTOR, '#play-store')
+
+    @property
+    def playstore_link_is_http_release(self):
+        href = self.selenium.find_element(
+            *self._playstore_locator
+        ).get_attribute('href')
+        return href == 'https://play.google.com/store/apps/details?id=org.mozilla.firefox#details-reviews'
```
Osmose: I think this should instead return the href, and the test itself should contain the URL you're asserting.

smoketests/pages/android_feedback_form.py
```diff
@@ -0,0 +1,153 @@
...
+        self.selenium.find_element(*self._sad_button_locator).click()
+        self.wait_for(self._moreinfo_text_locator)
+
+    # Moreinfo card
+    _moreinfo_text_locator = (By.CSS_SELECTOR, '#moreinfo #description')
+    _url_locator = (By.ID, 'id_url')
+    _submit_locator = (By.CSS_SELECTOR, '#moreinfo #form-submit-btn')
+    _checkbox_locator = (By.ID, 'last-checkbox')
+
+    def set_description_execute_script(self, text):
```
Osmose: I don't see this being used anywhere.

smoketests/pages/android_feedback_form.py
```diff
@@ -0,0 +1,153 @@
...
+        """
+        text = text.replace("'", "\\'").replace('"', '\\"')
+        self.selenium.execute_script("$('#description').val('" + text + "')")
+
+    def set_description(self, text):
+        desc = self.selenium.find_element(*self._moreinfo_text_locator)
+        desc.clear()
+        desc.send_keys(text)
+
+    def update_description(self, text):
```
Osmose: Also unused.

Osmose:
There's a pre-commit hook that lints the code for things like double quotes that I think might be useful to run: http://fjord.readthedocs.org/en/latest/conventions.html#git-pre-commit-hook

Fun fact, I can't actually get it to work on my machine for some odd reason. Pre-commit fails to install for me, and I've never been able to figure out why.

smoketests/tests/android/test_android_feedback.py
```diff
@@ -0,0 +1,102 @@
...
+import pytest
+from tests import TestCase
+from pages.dashboard import DashboardPage
+from pages.android_feedback_form import AndroidFeedbackFormPage
+
+
+class TestFeedback(TestCase):
+
+    @pytest.mark.nondestructive
+    def test_on_device_links(self, mozwebqa):
```
Osmose: This test and the one below it seem like multiple separate tests bundled into one. Each time you're calling `go_to_feedback_page` should really be a separate testcase.

smoketests/tests/android/test_android_feedback.py
```diff
@@ -0,0 +1,102 @@
...
+import time
+import pytest
+from tests import TestCase
+from pages.dashboard import DashboardPage
+from pages.android_feedback_form import AndroidFeedbackFormPage
+
+
+class TestFeedback(TestCase):
+
+    @pytest.mark.nondestructive
```
Osmose: I think all of these tests are considered destructive as they modify the database, but I might be wrong. @willkg will be able to confirm.

fjord/feedback/tests/test_views.py
```diff
@@ -193,6 +193,29 @@ def test_firefox_os_view_works_for_all_browsers(self):
         r = self.client.get(url, HTTP_USER_AGENT=ua)
         assert template_used(r, 'feedback/fxos_feedback.html')
 
+    def test_android_view(self):
+        """Android returns correct view"""
+        # Specifying fxos as the product in the url
```
Osmose: Comment is out of date and also kinda unnecessary.

fjord/feedback/tests/test_views.py
```diff
@@ -193,6 +193,29 @@ def test_firefox_os_view_works_for_all_browsers(self):
...
+            'happy': 0,
+            'description': u'Video plays backwards :(',
+            'url': u'http://mozilla.org/'
+        }
+        r1 = self.client.post(url, data, HTTP_USER_AGENT=ua)
+        self.assertRedirects(r1, reverse('thanks'))
+        # The below tests whether the feedback view function
+        # register_feedback_config() is working as it should
+        r2 = self.client.get(r1.url)
+        assert template_used(r2, 'feedback/android_thanks.html')
```
Osmose: Is there a reason you can't use that on `r1`?

fjord/feedback/tests/test_views.py
```diff
@@ -193,6 +193,29 @@ def test_firefox_os_view_works_for_all_browsers(self):
...
+        """Android returns correct view"""
+        # Specifying fxos as the product in the url
+        url = reverse('feedback', args=(u'android',))
+        r = self.client.get(url)
+        assert template_used(r, 'feedback/android_feedback.html')
+
+    def test_android_view_redirects_to_android_thanks(self):
+        """Sad feedback in Android redirects to proper thanks template"""
+        url = reverse('feedback', args=(u'android',))
+        ua = 'Mozilla/5.0 (Android; Tablet; rv:24.0) Gecko/24.0 Firefox/24.0'
```
Osmose: Does the User-Agent matter? I thought it was keyed off on the product slug.

Osmose:
This was a lot more understandable than I was expecting, and was pretty easy to follow. Nice work!

The smoketests could maybe use a bit more thorough coverage (like testing that the checkbox for including the URL works) but they're pretty close.

Drop a comment when you're ready for me to take another quick pass over this. Once I'm happy with it we'll have @willkg confirm that it won't break other stuff that I'm not aware of.

fjord/feedback/static/js/android_feedback.js
hoosteeno: Good call. I tested using a custom nightly mleibovic made for me, plus the console command in https://bugzilla.mozilla.org/show_bug.cgi?id=1235902#c8, and the event-constructed event works great.

fjord/feedback/tests/test_views.py
hoosteeno: I'm not sure I understand. I could clobber the response variable r1 on line 216 here and then test whether the thanks page is using the right template in the r1 response, but I thought this was more verbose. Don't have strong feelings about it so if there's a reason to clobber, I can do that.

Osmose: Ah, sorry, I forgot that the Django test client doesn't follow redirects by default. If you pass `follow=True` I _think_ you should be able to call both `assertRedirects` and `template_used` on one response, rather than making two requests.

If that doesn't work, then leaving this as it is is fine.

smoketests/tests/android/test_android_feedback.py
hoosteeno: Only the submit_sad modifies the DB. Will mark it as terribly destructive. :bomb: :boom:

hoosteeno:
@Osmose, thanks for good comments. I think the coverage in smoketests already address the checkbox -- see https://github.com/mozilla/fjord/pull/723/files#diff-3e75676286eecb6a2bf49bc3bca8e49bR114 and https://github.com/mozilla/fjord/pull/723/files#diff-3e75676286eecb6a2bf49bc3bca8e49bR132. If the URL was not sent after unchecking, the checkbox performed its most essential function.

More generally, I think the smoketests cover all the functionality in this form that differs from other, more standard feedback forms.

Osmose:
I'm an r+ on this, great work!

@willkg What do you think?

hoosteeno:
Just FYI, I took another look this morning and discovered a few non-breaking bugs in the variables at the top of some smoketests that would've perplexed maintainers. Pushed code to fix just now.

fjord/feedback/tests/test_views.py
willkg: We've got an older django-jinja that doesn't support the Django template rendering signals that you need when testing template_used. I wrote some shite code to sort of get around this, but it only shows the top-most template and probably won't work in this situation.

Pretty sure this would be true if we upgraded django-jinja to the latest, though.

PR #724, opened 2015-12-28
[bug 1235425] Remove "requires firefox" dead code

willkg:
This code is dead code--it's not being used or tested. This is just a
formal removal of the code so it's not crufting up the codebase
anymore.


willkg:
This is just removing dead code. It's pretty benign as far as code-removal goes. I'm going to self-r+ it.

PR #725, opened 2015-12-28
[bug 1235423] Refactor feedback_router

willkg:
This splits the picker view from the feedback_router code. Having said
that, the oldandroid code makes that continue to look a bit goofy.

This drops the requirement that we *only* show the Firefox OS form to
users of Firefox OS. With these code changes, if they got to
/feedback/, they'll now go to the picker. We don't get a lot of
Firefox OS feedback. As of 1.3, they have an in-product form that uses
the API anyhow.

This gets rid of the PRODUCT_OVERRIDES scaffolding which we haven't
used in ages and has very much overstayed its welcome.

The end result is a feedback_router that's a little more
straight-forward.

### DIFF
--- a/fjord/feedback/tests/test_views.py
+++ b/fjord/feedback/tests/test_views.py
@@ -21,11 +21,11 @@ class TestRedirectFeedback(TestCase):
 
     def test_happy_redirect(self):
         r = self.client.get(reverse('happy-redirect'))
-        self.assertRedirects(r, reverse('feedback') + '?happy=1')
+        self.assertRedirects(r, reverse('picker') + '?happy=1')
 
     def test_sad_redirect(self):
         r = self.client.get(reverse('sad-redirect'))
-        self.assertRedirects(r, reverse('feedback') + '?happy=0')
+        self.assertRedirects(r, reverse('picker') + '?happy=0')
 
 
 class TestFeedback(TestCase):
@@ -176,12 +176,6 @@ class TestFeedback(TestCase):
 
     def test_firefox_os_view(self):
         """Firefox OS returns correct view"""
-        # Firefox OS is the user agent
-        url = reverse('feedback')
-        ua = 'Mozilla/5.0 (Mobile; rv:18.0) Gecko/18.0 Firefox/18.0'
-        r = self.client.get(url, HTTP_USER_AGENT=ua)
-        assert template_used(r, 'feedback/fxos_feedback.html')
-
         # Specifying fxos as the product in the url
         url = reverse('feedback', args=(u'fxos',))
         r = self.client.get(url)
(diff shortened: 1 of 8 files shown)


willkg:
@hoosteeno What do you think of this? There's a lot of churn, but the architectural changes are pretty small. It's still if-block driven and not config driven. Though we could change that at this point. This doesn't address the Thank You step of the flow at all. I'm still mulling over options there.

hoosteeno:
I am not adequately steeped in django to r? this, but it looks like good work to me. I'll be happy to rebase on the final version of this and make sure the new android_feedback function works.

willkg:
Yeah--don't do any rebasing quite, yet. I've got a few more changes I want to make and then we'll see where we're at.

willkg:
^^^ That adds thank you view stuff, too. It's just like what you did in your first iteration. So your new android view code would look something like:

```
@register_product_view('android', 'feedback/thanks_android.html')
@csrf_protect
def android_feedback(request, locale=None, product=None, version=None, channel=None):
    """Android feedback form"""
    form = ResponseForm()

    if request.method == 'POST':
        return _handle_feedback_post(request, locale, product,
                                     version, channel)

    return render(request, 'feedback/android_feedback.html', {
        'form': form,
        'product': product,
        'TRUNCATE_LENGTH': TRUNCATE_LENGTH,
    })
```

I think that gets you everything you need requirements-wise. It puts all the details about this specific form in one place.

How does this strike you?

hoosteeno:
Strikes me as great. Thanks for working on this.

fjord/feedback/tests/test_views.py
```diff
@@ -176,12 +176,6 @@ def test_happy_prefill_in_querystring_is_ignored(self):
 
     def test_firefox_os_view(self):
         """Firefox OS returns correct view"""
-        # Firefox OS is the user agent
-        url = reverse('feedback')
-        ua = 'Mozilla/5.0 (Mobile; rv:18.0) Gecko/18.0 Firefox/18.0'
-        r = self.client.get(url, HTTP_USER_AGENT=ua)
-        assert template_used(r, 'feedback/fxos_feedback.html')
```
willkg: I ditched this feature. Basically, if you go to `/feedback/` with a Firefox OS phone, it would guess that you were using Firefox OS based on the user agent and then you'd get the Firefox OS feedback form. That's a terrible idea.

Now, it'll send you to the product picker and you can pick Firefox OS. It won't affect many people since Firefox 1.3+ uses an in-product feedback form and not the one hosted on Input.

willkg:
Nixing the WIP flag. I think this is better than what we had before.

Things to do:

1. give it another once-over
2. run it through the smoketests
3. land

willkg:
I ran the smoketests and they look good.

willkg:
Going to wait for Travis to give me another thumbs-up. Then I'll self-r+ it and merge it in.

willkg:
Travis is happy!

PR #726, opened 2015-12-29
Change some fields to use fuzzy values

willkg:
This makes it easier to have tests generate multiple surveys and
answers without hitting integrity issues.


willkg:
Travis is happy with this and it's really benign and only affects tests. I'm going to self-r+ it and land it.

PR #729, opened 2016-01-04
[bug 1236581] Fix Thank You view

willkg:
I goofed when I added support for product_config where the keys to
product slug were product slugs, but what we're storing in the
feedback_response table is the product db_name. Thus doing a lookup
using the product db_name would always fail. Oops!

This fixes that.


willkg:
Travis is green, so I'm going to self-r+ this.

PR #730, opened 2016-01-05
[bug 1129703] Switch gengo balance check to use mailing list

willkg:
This cleans up some code I wrote a while ago such that it now uses a
mailing list rather than mailing the admin when there are gengo
balance alerts.


willkg:
This is pretty benign and the tests cover the sending situations pretty well. I'm going to self-r+ this (last one--I promise!).

PR #731, opened 2016-01-12
[bug 1239017] Add URL pre-filling via URL

Osmose:
This allows you to create a URL that pre-fills the feedback URL.

    /feedback/?url=somewebsite.com

### DIFF
--- a/fjord/feedback/static/js/generic_feedback.js
+++ b/fjord/feedback/static/js/generic_feedback.js
@@ -160,5 +160,8 @@
                 selectHappySad(1);
             }
         }
+        if (qs.url) {
+            $('#id_url').val(qs.url).trigger('input');
+        }
     });
 }(jQuery, fjord, cards, remoteTroubleshooting, document, window));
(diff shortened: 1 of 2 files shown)


fjord/feedback/static/js/generic_feedback.js
```diff
@@ -160,5 +160,8 @@
                 selectHappySad(1);
             }
         }
+        if (qs.url) {
+            $('#id_url').val(qs.url);
```
willkg: Would we want to do any data validation/fixing?

Also, if we set the value, does that trigger the input event (or whatever it is that validates the value for this field)? If not, we should make sure that gets triggered.

Osmose: I think if we trigger the input event validation we don't need to validate before pre-filling, since the form will do the validation for us already. I will make it trigger the event.

Osmose:
Updated, got the smoketests running locally and they seem to be passing.

@willkg r?

willkg:
Looks good to me.
r_plus

PR #732, opened 2016-02-17
[bug 1215125] Fix whimsical locale in admin

willkg:
Django 1.8 changed trans_real.activate such that if you pass in a ''
language, then it does nothing. Since prefixer.locale is '' when
handling an /admin/ request, the request gets handled with the
previously handled request's activated translations.

This fixes that so that if prefixer.locale is '', then we use the
settings.LANGUAGE_CODE to activate the translations.

r?

### DIFF
--- a/fjord/base/middleware.py
+++ b/fjord/base/middleware.py
@@ -160,4 +160,7 @@ class LocaleURLMiddleware(object):
 
         request.path_info = '/' + prefixer.shortened_path
         request.locale = prefixer.locale
-        translation.activate(prefixer.locale)
+        # prefixer.locale can be '', but we need a real locale code to activate
+        # otherwise the request uses the previously handled request's
+        # translations.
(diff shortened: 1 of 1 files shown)


willkg:
This is kind of a pain in the ass to test locally because `fjord/wsgi.py` is total crap and `wsgi/playdoh.wsgi` is total crap and gunicorn needs a non-crap wsgi Python module that's specified as a Python module path and not a file path. Yay.

We can trivially test this on -stage by:

1. in a browser, log into -stage and go to the admin
2. in a different browser, go to https://input.allizom.org/fr/
3. refresh the admin

That causes the admin to show up as French for me. We can do any language--it's not specific to French. This might not always always work depending on whether -stage is handling other requests. But -stage is quiet enough that it works most/all of the time for me.

Osmose:
r+, nice fix!

willkg:
Woo hoo!

PR #733, opened 2016-02-17
Bug 1235417: Switch to git for storing translations.

Osmose:
I didn't touch `bin/update_site.py` because I assumed it was out-of-date as a deploy script, but I didn't want to just remove it. Maybe I should?

The process here is:
1. Merge this branch.
2. File a webops bug to replace the svn directory on staging with a git repo.
3. Once staging has the right directory, deploy this update.
4. Switch Pontoon's Input project over to the new git repo and test the entire process from submitting a Pontoon translation to deploying it to staging.
5. If things go well, repeat steps 2-4 on production.

@mythmon r?

### DIFF
--- a/bin/update/update.py
+++ b/bin/update/update.py
@@ -36,21 +36,21 @@ def update_product_details(ctx):
 
 @task
 def update_locales(ctx):
-    """Update a locale directory from SVN.
+    """Update a locale directory from Git.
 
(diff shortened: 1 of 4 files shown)


mythmon:
This seems good to me. r+

I'll let you handle merging and deploying this, since you'll probably need to do something tricky.

PR #736, opened 2016-03-10
[bug 1249881] Fix semantics for CheckMissingVotes health check.

Osmose:
See [this bugzilla comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1249881#c9) for a description of how the check is intended to work.

@mythmon r?


mythmon:
Ok, I think I understand what is going on here, and the code you wrote reflects the contents of the bug. It is pretty hard to reason about it all though. That's probably mostly because my grasp of SQL is fairly weak. The Python code you wrote looks solid.

r+

PR #737, opened 2016-04-14
[bug 1253423] Multiple bug fixes

hoosteeno:
Bugs addressed:
* Add proper SVGs, get rid of last_url checkbox and code handling last_url https://bugzilla.mozilla.org/show_bug.cgi?id=1253423
* Stop using a parameter to determine whether visitor is coming from fennec feedback prompt; start using a custom event fired by fennec: https://bugzilla.mozilla.org/show_bug.cgi?id=1256247

Also:
* smoketests updated


hoosteeno:
When this lands we should close...
* https://bugzilla.mozilla.org/show_bug.cgi?id=1256247
* https://bugzilla.mozilla.org/show_bug.cgi?id=1253423
* https://bugzilla.mozilla.org/show_bug.cgi?id=1225676
* (make sure someone else closes) https://bugzilla.mozilla.org/show_bug.cgi?id=1259186

Osmose:
Testing locally (both in the browser and on Fennec), I noticed a few things that I'm unsure about:

- After leaving positive feedback, I only see a thank you message, and no links to the Google play store. Is that expected?
- On Fennec, when the keyboard is open, the support bar at the bottom of the page isn't attached to the edges of the screen. This might be a pre-existing bug.

I'm unable to get my Nightly instance to open up my local Input in response to clicking the "Give feedback" button, so I can't really test the event. I'll leave testing that to you.

fjord/feedback/static/js/android_feedback.js
```diff
@@ -1,62 +1,51 @@
...
+            }
+            $('#play-store').attr('href', reviewUrl);
+        }
+        else {
+            // If not beta/release, hide CTA to review on play store
+            $('.play-store-cta').hide();
+        }
+
+        // Trigger input's generic_feeback.js handler
+        $('#happy-button').click();
```
Osmose: This is a very weird way of triggering the logic from generic_feedback.js. You should move the event handler into a function and call it here instead of simulating the click.

Osmose:
Assuming that the blank thank you page is expected (which I think it is based on the code, since I was viewing it in Firefox desktop), once you fix that simulated click thing I'm r+ on this.

One question: We want to wait until the event change gets uplifted to release to push this out, right? Or do we want to release this earlier?

Osmose:
Test failure is unrelated, the tests passed fine.

fjord/feedback/static/js/android_feedback.js
hoosteeno: I agree. My reasoning was to avoid refactoring code used by the entire card system for all products. "Simulating the click" really means calling the click event handler -- in the words of jQuery docs[0], "Execute all handlers and behaviors attached to the matched elements for the given event type". Which is actually what I want to do after my thing here. So I totally agree, this approach is unusual; but it seems legit to me. Am I off base?

[0] "click" is a shortcut for trigger: http://api.jquery.com/trigger/

hoosteeno:
The blank page is not great, but it is expected at this point. I think we should file a bug to give users a "next thing" on this page; anything would be better than what we have now, except requiring localizers to hurry up before launching this.

Osmose:
You'll want to rebase on master once https://github.com/mozilla/fjord/pull/739 lands to fix the tests.

fjord/feedback/static/js/android_feedback.js
Osmose: My main complaint is that this obfuscates what code is actually running. If I'm just looking at this file, in order to figure out what code is called when this line runs, I need to look through the _entire_ codebase to find every click handler that could possibly be registered to `#happy-button`. That might include more than just stuff registered by ID; if the tag is an anchor, a handler for all `<a>` tags might also run.

Also, this requires that you include a hidden tag in the dom just to trigger this. The core thing you want is to reuse a piece of code in two places, and factoring that code into a function is the standard and clearest way to do this. Introduce the DOM and events just obscures things even if it happens to result in less lines changed.

hoosteeno:
> On Fennec, when the keyboard is open, the support bar at the bottom of the page isn't attached to the edges of the screen. This might be a pre-existing bug.

✔️ Fixed.

> This is a very weird way of triggering the logic from generic_feedback.js. You should move the event handler into a function and call it here instead of simulating the click.

✔️ Fixed. I did it a different way (turns out it's easy to replicate the pieces of generic behavior this code needs just by using window.cards). I also wrapped this in a document ready, to mirror code structure in other files like this one.

> I was unable to test the event (tried using a nightly build with the commit that added it and Nightly's WebIDE to trigger the event) so I'm leaving the testing of that to you.

You can simulate all the conditions necessary to test things in-page in your desktop:
* To test that release/beta/nightly versions get the right behavior, use a URL like http://localhost:8000/en-US/feedback/android/46.0a/nightly or http://localhost:8000/en-US/feedback/android/46.0a/release
* To test that the fennec-fired event causes the right behavior, fire it yourself from the console: `var feedbackEvent = document.createEvent("Events"); feedbackEvent.initEvent("FeedbackPrompted", true, false); document.dispatchEvent(feedbackEvent);` 

If you really want to get simulate the whole thing in Fennec nightly...
* you need to be running a version of nightly with the feature included
* you need to change the config param "app.feedbackURL" to the URL of your locally running copy (e.g http://192.168.0.2:8000/feedback/android/46.0a/beta/)
* you need to simulate the feedback prompt by calling `Services.obs.notifyObservers(null, "Feedback:Show", null);` in the main process.

Osmose:
I still see the support bar issue with the keyboard open on Firefox for Android. Specifically this is on a Nexus 5X.

Still not a blocker. :D

Osmose:
r+ on the code. Nice job!

Could you rebase this on master? I just want to get Travis to pass the tests before merging.

Osmose:
Master's tests are passing again, rebase pleeeez

PR #738, opened 2016-04-20
Fix bug 1264916: Always pass a str to parse_qsl.

Osmose:
When a URL contains unicode, Django's request.get_full_path() returns
a unicode string with percent-encoded unicode strings. parse_qsl fails
to parse this properly, and returns a unicode string with bytes in it,
instead of codepoints. To fix this, we encode the URL in ASCII before
passing it to parse_qsl.

See http://www.lexev.org/en/2013/parse-url-which-chontains-unicode-query-using-urlp/ for a slightly more detailed description.

My local tests are borked so I'm hoping the automated tests will fare better. 

@mythmon r?

### DIFF
--- a/fjord/base/templatetags/jinja_helpers.py
+++ b/fjord/base/templatetags/jinja_helpers.py
@@ -38,7 +38,7 @@ def urlparams(url_, hash=None, **query):
 
     # Use dict(parse_qsl) so we don't get lists of values.
     q = url.query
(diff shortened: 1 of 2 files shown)


mythmon:
I'd write a test for this, to prove it works.

mythmon:
Merge conflicts!

Osmose:
@mythmon PR updated.

mythmon:
r+

PR #739, opened 2016-04-20
Fix requirement hashes.

Osmose:
PyPI is serving us zip files instead of tar.gz/tar.bz files, causing
peep to fail hash verification. I verified the hashes we had against
PyPI's tarballs, diffed their contents against the zipfiles to
ensure they hadn't changed, and then got the hashes of the zipfiles
and added them here.

See https://bitbucket.org/pypa/pypi/issues/436/peep-hash-failures-due-to-being-served-zip
for more info.


Osmose:
@mythmon r?

mythmon:
r+

